PR #8, opened 2019-08-18
python 3.6 compatibility

dhdaines:
Fix a couple of things so that the tests will run under Python 3.6.8 (and change the requirement in setup.py).  Needs testing on Windows!


roedoejet:
Thanks, @dhdaines! We'll have a look at this this coming week with @finguist's troublemaking Windows machine :)

PR #25, opened 2020-03-30
feat(lml): added raga g2p mapping

roedoejet:

* **Other information**:


roedoejet:
Hey Eric,

Thanks for this, as you noted, it's not actually an issue with g2p. G2P is performing the transduction correctly. \u02D0 is the correct character. The issue is in how ReadAlongs interprets the indices. Looking at the code in there, I guess it was written for a previous understanding of indices that assumed only one edge per input character (to the last character it maps to), but in g2p, there can be many edges between character nodes, so I added a filter in the dev.g2p branch (https://github.com/ReadAlongs/Studio/pull/25/commits/1cd3c79f6c443223ec010bab994530c6785dc62b).

When I run it with the above change, it appears to convert properly.

joanise:
Thanks @roedoejet ! That fixed the problem. It doesn't align, but we didn't really expect it to...

roedoejet:
> Thanks @roedoejet ! That fixed the problem. It doesn't align, but we didn't really expect it to...

Yea, at least the text is getting processed though! It'll be good test data for getting meaningful feedback from the decoder.

PR #35, opened 2020-04-19
Dev.crl

deltork:
updated CRL mapping


roedoejet:
This is awesome Del, thanks so much! I was going to try and get to this next week, but it's awesome you put the request in. I bumped down the minor version from 0.6 to 0.5, because we're leaving mapping updates to the date versioning, so the release on pypi will be something like 0.5.20200419 but that will happen automatically. I should document a versioning strategy though. I've pulled your changes into my dev branch and am merging with the master now, so you should be able to use it in readalongs! Thanks again for this!

PR #37, opened 2020-04-21
CRL completed!

deltork:
If it passes the tests, could you push the mapping to crj as well.


roedoejet:
Thanks Del! Is the mapping the same for crj as crl? How do you mean push it?

deltork:
> Thanks Del! Is the mapping the same for crj as crl? How do you mean push it?

yes we could just copy and paste the mappings from crl to crj

roedoejet:
Done: e116594

roedoejet:
And 3be6d7b, woops!

PR #41, opened 2020-04-27
Ojibwe and some changes to auto-generation

dhdaines:

* **Other information**:

I noticed when generating the mapping to English IPA that the settings for as_is and prevent_feeding (which is quite important for double-vowel orthography) and the ordering of rules from the input language weren't being preserved, which I think they should be as it is what a reasonable rule-writer would expect.  Also in general we were not saving prevent_feeding when writing config.yaml, so I fixed that.  I updated the unit tests to reflect this.


dhdaines:
I sort of see your point about as_is, but I think that especially now that we are preserving the order of the rules from the first mapping that we should also preserve the as_is flag.  Unless I'm mistaken the effect of this flag is on the order in which the rules are applied - and so if the rule writer expects the rules to be applied in a certain order in the first mapping, the derived rules in the generated mapping should also get applied in that order.

roedoejet:
Yes. `as_is` set to true applies the rules in the order they're declared. `as_is` set to false reverse sorts them by the length of the input string. The issue is just that supposing you have rules where the output changes length from the input, it might cause problems to evaluate the rules in the same order. But by switching `as_is` we might also cause problems. Either way I'll merge this now. Thanks again.

PR #50, opened 2020-08-06
fix(crk): correct mapping for ô -> oʊ

eddieantonio:
Fixes issues with alignment not recognizing "ô". The English IPA became o + ː which became "OW ː" in the aligner.


g2p/mappings/langs/generated/crk-ipa_to_eng-ipa.json
```diff
@@ -13,19 +7,19 @@
...
     {
         "in": "i\u02d0",
-        "out": "i",
+        "out": "e\u02d0",
         "context_before": "",
         "context_after": ""
     },
     {
-        "in": "o\u02d0",
-        "out": "o\u02d0",
```
eddieantonio, line 22: Ideally, the `crk-ipa` -> `eng-ipa` would know to convert **oː** to **oʊ**, but making this map straight from SRO to **oʊ** works too!

eddieantonio:
Update: so it still maps <ô> to <oː> but now the generated `crk-ipa` to `eng-ipa` mapping has `prevent_feeding: true`, which prevents a double long vowel situation happening. I added a regression test all the way to `eng-arpabet` (my use case was using this in readalongs), so this should be all fixed now!

Issue #51, opened 2020-08-06
Consider changing configuration names


roedoejet:
Certain config file settings are not very clear. For example: `as_is` if true, applies rules in the order they are declared, but if set to false, will reverse sort them based on the length of the inputs. Instead we could consider:

```yaml
rule_ordering: as-is
```

vs.

```yaml
rule_ordering: sorted
```

Or something like that.

roedoejet:
Fixed as of https://github.com/roedoejet/g2p/pull/52

PR #52, opened 2020-09-02
add config option: rule_ordering

eddieantonio:
This adds a config option called `rule_ordering`, which is a more self-descriptive alias of `as_is`.

Why?
-----

Say I am explaining how to write a mapping.

> Rule order matters!  This is how you can affect how `g2p` orders your rules:
>
>  - `as_is: true` means your rules are applied as written
>  - `as_is: false` means your rules are sorted such that the longest rules are applied first

I suggest changing the explanation this:

> - `rule_ordering: as-written` means the rules are applied in the order they are written in the mapping file
> - `rule_ordering: apply-longest-first` means that `g2p` will first sort these rules, such that rules with the longest input are applied first; this prevents them from taking part of unwanted feeding relations.

---

Addresses #51


joanise (commented):
I'm not keen on having two options that are synonyms of one-another, to be honest, but I understand the desire to rename this flag since `as_is` is not intuitive or self-documenting.

If we all agree `rule_ordering=...` is preferable, I would suggest converting all the existing mappings from `as_is` to `rule_ordering` and issuing a warning when the system sees `as_is` indicating it's deprecated though still supported. Or go one step further: convert all the rules and stop accepting `as_is` altogether, outputting instead an message telling the user how to convert correctly to the new syntax.

Furthermore, the documentation for `as_is` should be updated to say it means the same as the other option, and please use that one instead. And, the doc for `rule_ordering` should say "This was formerly done by setting `as_is=False`" or "`as_is=True`", respectively.

joanise:
Nice testing of the logging messages, I needed just that to test my changes on `g2p doctor`!

eddieantonio:
It went from 3 files modified to 42 files modified with all of the `as_is` changed to `rule_ordering`! :joy:

joanise (approved):
Nice work, this is much better, and I am happy to have this self-documenting `rule_ordering` name for the option. Good unit testing - the coverage went up a notch, which is always good to see! As far as I'm concerned, you can merge this in, as long as Aidan gives his go-ahead too.

roedoejet:
Fixes: https://github.com/roedoejet/g2p/issues/51

PR #54, opened 2020-09-04
g2p doctor improvements

joanise:
This PR brings a few improvements to the `g2p doctor` command.

 - `g2p doctor` without the `-m <lang>` flag will now process all mappings
 - the message printed when there is an error displays the whole rule now, to make it easier to figure out where the problem came from
 - that message also puts a caveat saying it's OK to ignore the warning if you know the character is ramapped elsewhere
 - the URL to panphon is printed just once

This PR also updates the help messages from all the `g2p` CLI commands, to better document what they do, how to use them, what the options mean.

This PR does not do the following item from the ClickUp task: "add support for optional flag `is_intermediate=true` to warn doctor the rhs is not final." The intent of that item is to search through the rules and see if the non-IPA characters are remapped elsewhere, but this turns out to be pretty complicated, because IPA validity is context sensitive. E.g., the vowel lengthening symbol that looks like a colon is valid IPA if it's after an IPA vowel, but not otherwise. So properly testing if something gets remapped is non-trivial, and I'm leaving for later at the moment.


joanise:
@eddieantonio @roedoejet I'm planning significant improvements to this PR, don't review it yet please. I'll let you know when I'm ready.

eddieantonio:
Sounds good, @joanise! Please @-mention me when ready 👍

joanise:
OK, @eddieantonio @roedoejet this PR is now ready to review.

g2p/tests/test_doctor.py
```diff
@@ -0,0 +1,25 @@
...
+from g2p.log import LOGGER
+from g2p.mappings.langs.utils import check_ipa_known_segs
+
+class DoctorTest(TestCase):
+    def setUp(self):
+        pass
+
+    def test_ipa_known_segs_fra(self):
+        with self.assertLogs(LOGGER, level='WARNING') as cm:
+            check_ipa_known_segs(["fra-ipa"])
```
eddieantonio, line 13: This is awesome! I didn't know that the `self.assertLogs()` method exists!

g2p/cli.py
```diff
@@ -35,20 +35,31 @@ def cli():
...
+@click.option('--out-dir', type=click.Path(exists=True, file_okay=False, dir_okay=True),
+    help='Output results in DIRECTORY instead of the global "generated" directory.')
+@click.option('--list-dummy', default=False, is_flag=True, help="List the dummy phone inventory.")
+@click.option('--dummy/--no-dummy', default=False, help="Generate dummy fallback mapping to minimalist phone inventory.")
+@click.option('--ipa/--no-ipa', default=False, help="Generate mapping from LANG-ipa to eng-ipa.")
 @click.argument('in_lang', type=click.Choice([x for x in LANGS_NETWORK.nodes if not is_ipa(x) and not is_xsampa(x)]))
-@cli.command(context_settings=CONTEXT_SETTINGS)
-def generate_mapping(in_lang, dummy, ipa, out_dir):
-    ''' Generate English mapping.
+@cli.command(context_settings=CONTEXT_SETTINGS, short_help="Generate English mapping.")
```
eddieantonio, line 44: Add a description, maybe with better wording that this:

```suggestion
@cli.command(context_settings=CONTEXT_SETTINGS, short_help="Generate English IPA or dummy mapping.")
```

eddieantonio (approved):
LGTM: just one command summary that could be improved!

g2p/tests/test_doctor.py
roedoejet: Yea! I didn't know about this either. I'm pretty sure all three of us have tested logs in three different ways. I think I like this the best though.

g2p/cli.py
```diff
@@ -7,7 +7,7 @@
 from networkx import draw, has_path
 
 from g2p.transducer import CompositeTransducer, Transducer
-from g2p.mappings.create_fallback_mapping import align_to_dummy_fallback
+from g2p.mappings.create_fallback_mapping import align_to_dummy_fallback, dummy_inventory
```
roedoejet, line 10:
```suggestion
from g2p.mappings.create_fallback_mapping import align_to_dummy_fallback, DUMMY_INVENTORY
```

roedoejet: I like that you pulled this out of the function, but let's [capitalize](https://www.python.org/dev/peps/pep-0008/#constants) it

g2p/cli.py
```diff
@@ -35,20 +35,31 @@ def cli():
...
     '''
-    if not ipa and not dummy:
+    if not ipa and not dummy and not list_dummy:
         click.echo('You have to choose to generate either an IPA-based mapping or a dummy fallback mapping. Check the docs for more information.')
     if out_dir and (os.path.exists(os.path.join(out_dir, 'config.yaml')) or os.path.exists(os.path.join(out_dir, 'config.yaml'))):
         click.echo(
             f'There is already a mapping config file in \'{out_dir}\' \nPlease choose another path.')
         return
+    if list_dummy:
+        print("Dummy phone inventory: {}".format(dummy_inventory))
```
roedoejet, line 62:
```suggestion
        print("Dummy phone inventory: {}".format(DUMMY_INVENTORY))
```

g2p/mappings/create_fallback_mapping.py
```diff
@@ -8,8 +8,9 @@
 from g2p.mappings.create_ipa_mapping import align_inventories
 from g2p.mappings.utils import is_ipa, unicode_escape
 
+dummy_inventory = ["ɑ", "i", "u", "t", "s", "n"]
```
roedoejet, line 11:
```suggestion
DUMMY_INVENTORY = ["ɑ", "i", "u", "t", "s", "n"]
```

roedoejet: There are two other places in the `align_to_dummy_fallback` where we'll have to rename `dummy_inventory` to `DUMMY_INVENTORY`

g2p/cli.py
```diff
@@ -114,23 +136,55 @@ def convert(in_lang, out_lang, input_text, path, debugger):
...
-    '''
+
+        You can list available mappings with --list-all or --list-ipa, or by visiting
+        http://g2p-studio.herokuapp.com/api/v1/langs .
+    """
+    if list_all or list_ipa:
+        out_langs = sorted(set([x["out_lang"] for x in MAPPINGS_AVAILABLE]))
+        if list_ipa:
+            out_langs = [x for x in out_langs if is_ipa(x)]
+        print("Specifying an output language will check all mappings into that language:")
```
roedoejet, line 161:
```suggestion
        LOGGER.info("Specifying an output language will check all mappings into that language:\n")
```

g2p/cli.py
```diff
@@ -114,23 +136,55 @@ def convert(in_lang, out_lang, input_text, path, debugger):
...
+        for m in out_langs:
+            print(f"{m}:", end="")
+            for n in sorted([x["in_lang"] for x in MAPPINGS_AVAILABLE if x["out_lang"] == m]):
+                print(f" {n}->{m}", end="")
```
roedoejet, line 165:
```suggestion
                print(f" {n}->{m}\n\t", end="")
```

roedoejet: I think this makes it a bit more readable. 
image

roedoejet (changes requested):
This is great @joanise! I can see this being very useful. I just capitalized the dummy_inventory constant and changed the formatting of the list-ipa flag output, but this is ready to merge. Note, there are a few places in the `align_to_dummy_fallback` function that will have to be renamed as well to uppercase. If you accept these suggestions you'll have to commit those as well, or I can make the commit later before merging instead, as you wish.

Thanks again!

roedoejet:
One other thing we'll need to do, is have a look at the output of `g2p doctor` and edit the mappings to accept or fix places that doctor complains.

g2p/tests/test_doctor.py
joanise: I was going to replicate what Eddie did in his PR, but I also did a web search and found on StackOverflow another complicated method, and then a second answer lower down in the thread showed this solution, which is implemented directly for this! Way easier! I'm glad I found it. It's apparently relatively recent.

g2p/mappings/create_fallback_mapping.py
joanise: More Python conventions I'm not familiar with yet. Sure, I'm happy to follow PEP8 on this too. I wondered if flake8 would have caught this, but to my surprise I just checked and it doesn't complain. Questions on StackOverflow suggest pylint would complain, but when I try it doesn't. 
No matter, this is clearly a constant and should be upper cased, I will accept the change.

g2p/cli.py
joanise: Agreed, this is much easier to read.

joanise:
@roedoejet Thanks for the detailed review. Indeed, we'll have to apply the fixes suggested, although when I looked at them most don't need changing: most messages are about temporary symbols that do indeed get resolved, and since we don't check for that yet (I analyzed this a bit, and it won't be trivial, which is why I didn't do it right away), we can't clean this output yet.

g2p/cli.py
joanise: @roedoejet I touched up the output a bit. As it was, it looked nice for `--list-ipa`, but not so much for `--list-all`. I'm now using spaces instead of tabs, and counting how many I need to line up each list under its out_lang heading.
Finicky detail I unreasonably care about: I also suppressed the extra whitespace on the blank line between lists.

PR #57, opened 2020-09-15
Fin

finguist:

- [X] Mapping files are added in `g2p/mappings/langs`
- [X] Mapping is either added to an existing folder or a new folder has been added
- [X] Language folder and files use appropriate [ISO 639-3 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-3_codes)
- [X] Config.yaml file includes all author names, and settings necessary
- [X] Please add some test data in `g2p/tests/public/data`. The added file should be a csv/tsv/psv file and each row should have the format `[input_mapping_code,output_mapping_code,input_string,output_string]`

* **Other information**:


roedoejet:
Hey @finguist thanks for the PR!

The PR isn't passing the tests though, it looks like maybe you haven't pulled from the master branch in a while. Can you pull and merge from the master branch? Also, what does tmd stand for? Turtle Mountain something? Can you provide a bit more information in either the config or a readme (in the `crg` folder) about what these are for? Is it possible to have a normalization mapping between these two systems? It would be nice to have a path from just `crg` to `eng-arpabet` if possible.

Thanks for the work on this Fin! It'll be a great addition to the repo!

roedoejet:
Oh! Also, by the way, the main repo has introduced a breaking change on one of the config variables, as of https://github.com/roedoejet/g2p/pull/52. We renamed `as_is` in the config to `rule_ordering`.

These are the equivalent settings:

`as_is: true` is equal to `rule_ordering: as-written`
`as_is: false` is equal to `rule_ordering: apply-longest-first`

PR #58, opened 2020-09-16
added w final

shankhalika-katya:
Added a missing syllabic


roedoejet:
good catch, thanks @shankhalika-katya !

PR #62, opened 2020-10-05
Fix: generate-mapping should output rule_ordering, not as_is

joanise:
`g2p generate-mapping --ipa <lang>` should also output "rule_ordering", not "as_is".


roedoejet (approved):
Thanks @joanise ! I forgot about this!

PR #64, opened 2020-10-22
fix(crk): change affricate to English-compatible IPA

eddieantonio:
\<c\> in Cree SRO represents the phoneme /t͡s/ can be realized as [t͡s ~ t͡ʃ], sometimes with notable voicing; however, when mapping to English ARPABET, `t͡s` turns into `T S͡` which then fails in downstream applications. This PR changes the mapping to `tʃ`.


roedoejet:
Hey @eddieantonio, thanks. Yeah, this actually is a result of the generated mapping for crk-ipa -> eng-ipa inheriting `prevent_feeding=true` from the `crk-no-symbols -> crk-ipa` mapping. Because there is a (kind of meaningless) rule that was generated for `t-> t` and because the config was set to prevent feeding, it was converting t into an 'intermediate' representation in the Unicode PUA, and then the rule that removes the breve wasn't applying.

PR #65, opened 2020-10-28
fix(g2p update): errors in mappings now produce user-meaningful error…

joanise:
fix(g2p update): errors in mappings now produce user-meaningful error messages

This is accomplished by have exceptions.MalformedMapping take a message
argument that gets printed after the generic information.

also:
 - Mappings in .csv/.tsv format are now allowed to have blank lines, which are
   simply ignored.
 - When a line in a .csv/.tsv file has only one field, an error message
   explicitly says the "out" field is required, via exceptions.MalformedMapping.

Blank lines in .csv/.tsv files is what triggered the current commit: previously,
a blank line accidentally added to the file would cause an index error saying
entry[0] had an index out of range, which was not informative.

### DIFF
--- a/g2p/exceptions.py
+++ b/g2p/exceptions.py
@@ -36,8 +36,11 @@ class InvalidNormalization(CommandLineError):
         ))
 
 class MalformedMapping(CommandLineError):
-    def __init__(self):
-        pass
+    def __init__(self, message=""):
+        if message:
+            self.message = "\n\n" + message
+        else:
+            self.message = ""
     
     def __str__(self):
         return self.render((
@@ -46,6 +49,7 @@ class MalformedMapping(CommandLineError):
             'You might be missing some keywords or \n'
             'Not all of the input and output pairs in your mapping have values for "in" and "out"\n'
             'Please refer to the documentation and fix your mapping.'
+            + self.message
         ))
 
 class MalformedLookup(CommandLineError):
--- a/g2p/mappings/__init__.py
+++ b/g2p/mappings/__init__.py
@@ -87,12 +87,12 @@ class Mapping():
                 abbreviations)
         # Handle user-supplied list
         if isinstance(mapping, list):
-            self.mapping = validate(mapping)
+            self.mapping = validate(mapping, path="user-supplied mapping")
         elif isinstance(mapping, str) and (mapping.endswith('yaml') or mapping.endswith('yml')):
             loaded_config = load_mapping_from_path(mapping)
             self.process_loaded_config(loaded_config)
         elif isinstance(mapping, str):
-            self.mapping = validate(load_from_file(mapping))
+            self.mapping = validate(load_from_file(mapping), path=mapping)
         else:
             if "in_lang" in self.kwargs and "out_lang" in self.kwargs:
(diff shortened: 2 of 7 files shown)


joanise:
Initial PR did not pass coverage test: coverage had gone down a tad.
New commits add test cases for invalid mappings to exercise the changes in the code.

I needed PortageNamedTemporaryFile for these tests, so I copied it from readalongs/tempfile.py, because I don't want to create a test we know will be broken on Windows. If we're OK with tempfile.py existing in g2p, we could strip it out of readalongs and use it from g2p instead, so we only have it in one place. But that's a funny dependency to create: readalongs depending on g2p utility code is a bit weird, so we could also keep it in both places, although I don't like that very much - I'd rather factor identical code out to a single copy.

roedoejet:
Thanks @joanis, don't you mean `PortableNamedTemporaryFile` - you've worked on Portage for too long! 😄 

There are a couple issues here. The tempfile is using the readalongs logger which is causing Travis to fail, as it should. I agree it wouldn't be a good idea to have readalongs import this from g2p. But, more than that, I'm not sure I understand why this utility class is needed at all here? Maybe we can have a conversation about it.

joanise:
I should have noticed the readalongs logger, but I guess when I ran the tests on my machine, they passed since I have it installed. Travis was quite slow to start yesterday, things staying in the queue a long time, so I didn't see it fail before I finished my work day.

I only need the tempfile for my unit testing, so I could just rewrite the code to not need it. It will make it longer, but that's not really a big issue.

The reason I didn't want to use to regular tempfile is that I like having the actually temporary file cleaned up when the class exists instead of when you close it, because the similarly simple code using `tempfile.NamedTemporaryFile` works on Mac and Linux but not on Windows.

Anyway, for just a few test cases, I'll move the logic to to `test_mapping.py`, where I need it, and remove the extra file. Let's close this PR and I'll create a cleaner one without the unnecessary files in its history.

I do want to add the improved error messages, that came from Fin not knowing what to do with the error message `g2p update` gave her when she had spurious whitespace in her `.csv` mapping file. So I'll create a clean PR with just that work.

joanise:
Re `Portable` vs `Portage`, you're right, I've worked on the Portage project for so many years, my fingers just automatically complete anything that starts with "Port" into "Portage". It's funny most of the time...

joanise:
OK, adding `tempfile.py` was just silly, given how trivial the alternative code is. Now I remember the reason for 
`PortableNamedTemporaryFile` in readalongs: in that case, we create the tempfile somewhere, and pass it out in a `return` statement, and we need the deletion to happen when the calling environment is closed, not the creating environment. That's why I needed to create that class to make that code work on Windows.

In this `g2p` test case, I'm still in the same scope so it's trivial to use `os.unlink()`. So, my apologies for the noise...

joanise:
PS: I love working with code reviews. We didn't do that in Portage, but each time we do it for g2p and readalongs, I find it highly valuable. We produce better code this way!

roedoejet:
Yea! I agree! Thanks for explaining this, I'm glad we can use a simpler solution here.

PR #66, opened 2020-10-28
chore: travis should run all tests, not just dev

joanise:
This PR is just to update the travis configuration to run all tests, not just `dev`. Right now, that makes no difference, but when tests are added they should automatically be included in the travis runs even if they are not added to a set that is part of dev.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_script:
 
 # commands to run the testing suite. if any of these fail, travic lets us know
 script: 
-  - cd /home/travis/build/roedoejet/g2p && coverage run --source g2p run_tests.py dev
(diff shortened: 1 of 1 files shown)


roedoejet:
Thanks @joanise - I'm not sure I'd like to do this though. Because the 'all' test suite just uses the `unittest.TestLoader.discover` method, we wouldn't have any control over which tests were run or not. So, if somebody is working on some tests that aren't finished yet, it will run them anyways and Travis might report a false failure for our tests. Of course we run the risk of someone forgetting to add the tests to the dev suite once they're complete, but at least with that we have control over it.

joanise:
I think tests in development that don't pass yet belong in branches, not in master, but I guess if you prefer to keep it the way it was, I don't mind. I suppose coveralls will tell us when we forgot to add a test to dev, since coverage will go down instead of up.

PR #67, opened 2020-10-29
Dev.malformed mappings take2

joanise:
- improved message when `g2p update` fails because of mapping errors.
- unit testing without unnecessary additional code.


dhdaines (approved):
This all looks good (assuming all the tests pass)

PR #72, opened 2020-11-13
Dev.cli scanner

saltyseadawg:
add scan cli command and data for test
- check n file against target language for any unmapped characters by comparing against the set of input characters that appear in all target lang mappings 
- currently fails test (ex. é appears as unmapped in cases where acute accent is transformed separately)


joanise:
The build fails at this point. We need to review the test suites and fix them before we can merge this code in.

joanise:
Right, just looked at the code, and it seems like it's the unit test case that catches an error with my French g2p that causes the unit tests to fail. I'll try to review my French g2p to fix this.

joanise:
I've modified the unit testing so that this PR no longer depends on my fixing my fra g2p.
However, there is still an issue with normalization that must be fixed before merging this PR: `g2p convert` applies NFD/NFC normalization (as per each mapping's configuration) before doing conversion, so that "é" and "e" + \u0301 are seen as the same.
`g2p scan` needs to reproduce that preprocessing to be accurate, especially for languages with lots of accented characters that can be normalized in more than one way.

joanise (approved):
Great work @saltyseadawg thanks for implementing scan and making the changes we suggested. This will be a useful tool in g2p.

PR #73, opened 2020-11-26
fix(g2p doctor): --list-* options should print langs sorted

joanise:
Tiny PR, I know, to make `g2p doctor --list-all` output languages in alphabetical order and thus make the output a lot more legible.


roedoejet (approved):
Hehe, what a cute little PR @joanise - yes. I approve 😄

PR #75, opened 2020-12-07
-norm to -equiv

finguist:
Removing -norm from the `config.yaml` files so the user sees -equiv.

### DIFF
--- a/g2p/mappings/langs/crj/config.yaml
+++ b/g2p/mappings/langs/crj/config.yaml
@@ -3,7 +3,7 @@
 mappings:
   - display_name: Northern East Cree Normalization
     in_lang: crj
-    out_lang: crj-norm
+    out_lang: crj-equiv
     type: mapping
     authors:
       - Delasie Torkornoo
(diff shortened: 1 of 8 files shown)


g2p/mappings/langs/crj/config.yaml
```diff
@@ -3,7 +3,7 @@
 mappings:
   - display_name: Northern East Cree Normalization
```
joanise, line 4: I think I would change the display name for each of these to "lang_name equivalency" or "lang_name character equivalency". You probably want to get rid of the word Normalization in the display names too.

I remember we agreed on the short form "equiv", did we agree on a long form?

joanise (changes requested):
Did you run the test suite on your machine? It fails on Travis.

It looks like you didn't run `g2p update`, or maybe just forgot to commit the changed files. That's needed to make the change effective.

The is one step further I think is worth doing for this, and it is to rename all the `tli_norm.csv`, etc, files, to `tli_equiv.csv`, which will further encourage future `g2p` contributors to use "equiv" in both file names and "lang" names.

PR #76, opened 2020-12-07
feat(mic): added mi'kmaq mapping

roedoejet:

* **Other information**:


roedoejet:
I just decided to go for it and do this - it appears to work. Apparently voicing is allophonic, so this is a kind of hybrid somewhere in between phonemic/phonetic detail. Basically just phonemic + voicing. Looks like there are also a number of orthographies, but this is the most used apparently.

joanise:
Cool! That was quick. I'll take a look this week, between sessions at COLING.

g2p/mappings/langs/mic/mic_to_ipa.json
```diff
@@ -0,0 +1,154 @@
...
+      "out": "x"
+   },
+   {
+      "in": "y",
+      "out": "j"
+   },
+   {
+      "in": "p",
+      "out": "b",
+      "context_before": "SONORANT",
```
joanise, line 41: I believe, with the way `context_before` is implemented, this context will be matched even if `ː` appears in isolation. I.e., `bːp` would get mapped to `bːb`. I think that's OK, because that's certainly not proper Mi'kmaq input or ipa intermediate, but it's best to be aware of it.

g2p/mappings/langs/mic/mic_to_ipa.json
```diff
@@ -0,0 +1,154 @@
...
+   },
+   {
+      "in": "^q",
+      "out": "əq",
+      "context_after": "CONSONANT"
+   },
+   {
+      "in": "^s",
+      "out": "əs",
+      "context_after": "CONSONANT"
```
joanise, line 112: Simple solution. I'm going to assume these vocalic consonants are all correct.

g2p/mappings/langs/mic/mic_to_ipa.json
```diff
@@ -0,0 +1,154 @@
...
+   },
+   {
+      "in": "^p",
+      "out": "əp",
+      "context_after": "CONSONANT"
+   },
+   {
+      "in": "^q",
+      "out": "əq",
+      "context_after": "CONSONANT"
```
joanise, line 107: This won't work, as you have q->x earlier in the mapping.

g2p/mappings/langs/mic/mic_to_ipa.json
```diff
@@ -0,0 +1,154 @@
...
+   },
+   {
+      "in": "p",
+      "out": "pə",
+      "context_after": "(CONSONANT)$"
+   },
+   {
+      "in": "q",
+      "out": "qə",
+      "context_after": "(CONSONANT)$"
```
joanise, line 142: Also won't work, running after q->x

joanise (changes requested):
Quickly done!

See various comments attached to the code.

More global comment:

I think there's a problem with all the ^C->eC followed C->Ce with context_after=CONSONANT in both cases, which I think will make ^C end up as eCe (read all these e's as schwas). Maybe instead, C->Ce followed by ^Ce->eC?

And we have the usual problem with ^, which matches the beginning of the line, not the beginning of the word... Maybe we need to change `g2p convert` to tokenize on spaces before applying the mapping, then `^` and `$` would match on all word boundaries, even within a line.

g2p/mappings/langs/mic/mic_to_ipa.json
roedoejet: > Quickly done!
> 
> See various comments attached to the code.
> 
> More global comment:
> 
> I think there's a problem with all the ^C->eC followed C->Ce with context_after=CONSONANT in both cases, which I think will make ^C end up as eCe (read all these e's as schwas). Maybe instead, C->Ce followed by ^Ce->eC?
> 
> And we have the usual problem with ^, which matches the beginning of the line, not the beginning of the word... Maybe we need to change `g2p convert` to tokenize on spaces before applying the mapping, then `^` and `$` would match on all word boundaries, even within a line.

The point about ^C->əC followed C->Cə is good except the latter rule is only turning C->Cə when followed by a word-final consonant, so it currently predicts sq -> əsəx (after spirantization of the /q/). I suppose you're saying it might be səx, but I'm not actually sure. səx, əsx and əsəx all seem plausible. I've re-ordered the rules as you propose though and I suppose I'll keep my ears out for these kinds of words!

The way `context_before` is implemented is to make a lookbehind by joining all of the characters in the SONORANT variable. The fully expanded regex for the rule above looks like this: `((?<=iː|aː|eː|oː|uː)|(?<=i|a|e|u|o|y|m|n|l))p(?=i|a|e|u|o|iː|aː|eː|oː|uː|y|m|n|l)` the single and two-character sequences are automatically separated into different groups because lookbehinds have to be fixed width. Anyways, long story short, `bːp` won't map to `bːb`.

g2p/mappings/langs/mic/mic_to_ipa.json
roedoejet: Thanks for these Eric, you're totally right! I used the visual programming rule generator to generate based on the CONSONANT variable but these are redundant after the q->x rule as you say.

roedoejet:
Thanks for these really helpful comments @joanise !! Very helpful indeed.

roedoejet:
@joanise - I believe I addressed your requested changes. I also gave it a quick run through with some of the data here: http://www.firstnationhelp.com/ali/ppmbook/ and alignment worked with this story. Let me know if anything else is needed otherwise feel free to approve and merge! Thanks!

joanise:
Looks good, as far as I can tell. I would like more examples in the unit testing, but that's not essential.

joanise (approved):

PR #77, opened 2020-12-08
fully switched user facing norm to equiv in config files and filenames

finguist:
I ran g2p update this time, as well as the test suite which passed. I also incorporated suggestions from @joanise :)


joanise:
Just to make the link for later retrieval, this PR is an update of PR #75

joanise:
This looks good to me, I think we should be able to merge as is. I'd like @roedoejet 's review too, just in case I missed something.

Aidan, for some context, we are trying to remove the word "normalization" since it has negative connotations. We agreed on "character equivalencies" or "-equiv" to replace "-norm" in mapping in/out lang names, and in file names.

This approach should be documented somewhere, which it is not yet: something to the effect that "When languages have different conventions for typing the same letter using different characters, we consider them to be equivalent characters. -equiv mappings are meant to convert input text from the set of equivalent characters to the one equivalent character that is recognized by the g2p mapping for that language." And refer to the several `*-equiv.*` files in `mappings/langs/*`.

joanise (commented):
Approved as far as I'm concerned, pending Aidan's approval.

roedoejet (approved):
I think this is a good solution. Thanks @finguist and @joanise !

PR #80, opened 2021-01-26
Dev.tce debug

shankhalika-katya:
resolving unicode issues with tce and changing generated file from NFC to NFD


roedoejet:
Thank you @shankhalika-katya !

PR #83, opened 2021-02-16
Patch.scan

saltyseadawg:
fix scan command and add more scan tests
- exclude mappings for converting between different IPAs from check 
- fix logic for checking case sensitivity
- add tests for case sensitivity


joanise (approved):
Looks good, thanks for improving scan!

PR #85, opened 2021-02-16
Patch for kwk-napa to kwk-ipa

finguist:
Added one missing line to the csv file


joanise (approved):
Thanks for the quick fix!

PR #87, opened 2021-02-22
Dev.str

shankhalika-katya:
added str-equiv 
changed config to reflect str-equiv
switched the order of the rules on ln123 and ln127 in str-to-ipa to ensure proper conversion of Ś to eng-arpabet
patch in str-ipa_to_eng-ipa to ensure e: is converted properly to eng-arpabet


g2p/mappings/langs/generated/str-ipa_to_eng-ipa.json
```diff
@@ -8,11 +8,11 @@
         "out": "dz"
     },
     {
-        "in": "e",
-        "out": "e\u02d0"
+        "in": "e\u02d0",
```
roedoejet, line 11: What is the purpose of transducing `e\u02d0 -> e` only to transduce it back from `e -> e\u02d0` in the next rule? It looks like you're not preventing feeding in the config and the rules are applied as-is, so are these two rules needed at all?

roedoejet: Oh, this is for input of `e\u02d0` from the previous mapping - I see. You're trying to prevent `eːː`?

g2p/mappings/langs/str/str_equiv.json
```diff
@@ -0,0 +1,78 @@
...
+        "in": "Á",
+        "out": "A\u0301"
+    },
+    {
+        "in": "Ⱥ",
+        "out": "Ⱥ"
+    },
+    {
+        "in": "Ć",
+        "out": "C\u0301"
```
roedoejet, line 12: Some of your rules are going from NFC to NFD (like A or C), and some are going from NFD to NFC (like I or S) - why is that? Is that what the keyboards do?

roedoejet (changes requested):
Hey @shankhalika-katya, thanks a lot for this! I realize I wasn't at the last meeting so maybe I'm just out of the loop, but if you could answer the questions I posted that'd be great - thank you!

g2p/mappings/langs/generated/str-ipa_to_eng-ipa.json
shankhalika-katya: yes exactly! It's a bit of a hack-y fix I know, but when we went over it with Eric, he said the issue was in the generate-mapping function and that this was a good enough fix to just get str working until the bug gets resolved in the actual function

g2p/mappings/langs/str/str_equiv.json
shankhalika-katya: yeah, I just kept the NFC/NFD configuration of the output characters as they already were in str-to-ipa, so this sometimes meant I was going to NFC and sometimes to NFD. Generally the letters like Ⱥ were NFD originally, and some of the ones like Í were NFC.

roedoejet: Interesting - is that confirmed from anything? Like, are we sure str-to-ipa is correct?

shankhalika-katya: Yes, I think so, because in both the SENĆOŦEN keyboards I'm using, the characters arent consistently NFC or NFD. theyre mostly NFC, but W̱ and X̱ are always NFD. However I double-checked now and noticed some rules that would normally be in the equiv are in the ipa-mapping itself (e.g. the rules on line7 and line11), so i deleted the redundancy in the equiv just in my sandbox and it runs alright on my end. Should i push that change, or should i move the rules out of str-ipa into the equiv?

roedoejet: > Yes, I think so, because in both the SENĆOŦEN keyboards I'm using, the characters arent consistently NFC or NFD. theyre mostly NFC, but W̱ and X̱ are always NFD. However I double-checked now and noticed some rules that would normally be in the equiv are in the ipa-mapping itself (e.g. the rules on line7 and line11), so i deleted the redundancy in the equiv just in my sandbox and it runs alright on my end. Should i push that change, or should i move the rules out of str-ipa into the equiv?

Good question - I think it would actually be best to move the normalization/equivalency parts of the ipa mapping to the equiv mapping, just to be more consistent with the purpose of each mapping. Can you do that and push the change? I'll merge it after that. Thanks @shankhalika-katya !

shankhalika-katya: done! thanks @roedoejet !

roedoejet: That was fast! Thanks @shankhalika-katya !

roedoejet (approved):

PR #89, opened 2021-03-12
Make the generated mapping files human readable

joanise:
By turning off the `ensure_ascii` option to `json.dump`, we can get generated mappings in `mappings/langs/generated` to use the Unicode characters themselves instead of their `\u` escapes, which will make the mappings much easier to read, as long as we're using a Unicode-compatible editor (which, really, we should always do!)

Code is changed to generate future mappings in this human readable fashion.

All existing generated mappings were updated with this script:
```
import sys
import json
input_json = json.loads(sys.stdin.read())
sys.stdout.write(json.dumps(input_json, indent=4, ensure_ascii=False))
```


littell:
Yes, agreed; this will make human review a lot easier.

littell (approved):

PR #90, opened 2021-03-12
new release

joanise:
Publish the g2p changes from the last three months to PyPi.


roedoejet (approved):
Thanks Eric, we should probably schedule this to happen every month or so? What do you think?

joanise:
Yeah, it would make sense to do it regularly. Monthly sounds reasonable.

joanise:
Thanks for merging!

PR #91, opened 2021-03-16
Dev.workflow: update the workflow to only publish from release branch

joanise:
Currently, the `pythonpublish.yml` workflow is triggered on
 - updating the `release` branch
 - pushing a tag called `release`
 - pushing a tag called `build`
 - pushing a tag matching `v*.*.*`

The workflow we want is to publish only when we update the `release` branch, which this PR does.

Also: link to PyPI repo from the README.


roedoejet (approved):
Looks good to me - thanks Eric!

PR #92, opened 2021-03-16
Automatically tag the repo on release

joanise:
This PR changes the `pythonpublish.yml` workflow to tag the repo with the version just created automatically.
I tested this reasonably well, in this branch, with a version that echoes the tag command instead of running it. And I tested the tagged process in a different GitHub repo [Just-testing](https://github.com/joanise/just-testing) where I successfully tagged the repo at each commit I pushed in a similar workflow, so I'm fairly confident this will work. But, we'll only know for sure when we do the next g2p release...

Risk: not too much. I put the tag steps at the end, so if it fails, it won't prevent the rest of the workflow from happening. So I think the only risk is that the next release might or might not actually get tagged.

Caveat: it's a bit flaky getting `$VERSION` by parsing the filenames in the `dist/` directory, but we don't save the full version anywhere during the build, except in the wheel and tarball file names. Maybe we should also put the version somewhere client code can see e.g., possibly by changing `setup.py` so that `python setup.py sdist bdist_wheel` would update `_version.py` with the full version number including the build number?

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -6,7 +6,7 @@ on:
   release:
     types: [created]
   push:
-    branches: [ dev.workflow_auto_tag ]
+    branches: [ release ]
 jobs:
   deploy:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


roedoejet (approved):
Hey Eric, 

This looked really good! I ended up switching it around a little bit though, here's what I did:

1. I removed the trigger for new releases since we'll be creating new releases here
2. I used the official GitHub action for tagging - this has two benefits. The first, is that the commit will appear to be done from the GitHub Action bot which is more appropriate than from me. The second is that it will automatically create a change log for any commits since the last release that were done with the conventional commit style.
3. I also added an official release action instead of just publishing the tag. This means that there will be downloadable wheels and zip files of our releases along with changelog information for each release.

Thanks again for getting the ball rolling on this!

joanise:
Well, you're using some magic I'm not familiar with here, but if your solution works, it's certainly tidier than mine. Obviously, we'll have to put flask at 1.0.2 before this branch can pass unit testing, maybe we should just merge master into here and undo Flash 1.0.2 once that problem is fixed.

joanise:
Actually, I propose we rebase this PR's branch on top of master and do a forced push, it's give us a much tidier history in the end, and it would pass CI this way.

PR #95, opened 2021-03-23
Dev.haa

shankhalika-katya:

- [x] You agree to license your contribution under the same license as this project (see [LICENSE](https://github.com/roedoejet/g2p/blob/master/LICENSE) file).

* **Other information**:


g2p/tests/public/data/haa.csv
```diff
@@ -0,0 +1,12 @@
+haa,haa-ipa,tth’ą̂ą̈̌,tθ'æ̃ɑ̃
+haa,haa-ipa,nhdhthtthddhwhshzhchch’yhkhghh'
+haa,haa-ipa,̥ndθtθʰtθ̥wʃʒtʃʰtʃ'jxɣhʔ
```
joanise, line 3: These two test cases only have 3 fields each, there's a comma missing between the input and the reference.

joanise (changes requested):
There was a merge conflict that `git merge` should have reported in `generated/config.yaml`, but instead it created an incorrect merge. I have fixed this.

There is also a problem with `tests/public/data/haa.csv`: two lines are missing a comma between the input and the reference, or maybe they're missing the reference, I'm not sure. Please check.

There is a third and more complicated problem: my tokenizer is compatible with two hops from lang to lang-ipa, but not yet with three hops from lang to lang-ipa. I'll have to fix that before the `haa.csv` test suite can pass completely.

joanise:
@shankhalika-katya I'll let you fix the `haa.csv` file. You'll know it's fine when running `./test_langs.py` in `g2p/tests` succeeds and outputs no warnings. Please fetch my extra commit on your `dev.haa` branch first.

I'll fix the other issues that are blocking this PR.

The other two PRs are blocked by the same problems as this one, so I'll want to fix this PR first and then deal with the other two.

g2p/tests/public/data/haa.csv
shankhalika-katya: fixed!

joanise (approved):

joanise:
Thanks @shankhalika-katya the fixed test suite. It works now.
Meanwhile, I have just fixed the tokenizer to support three hop settings like the one you needed for haa, so we can now merge this PR.

PR #101, opened 2021-04-12
Dev.all delimiters

joanise:
With my tokenizing transducer, the extra space added after words in eng-arpabet got lost.
This patch puts them back in, at the cost of having a spurious space at the end of all eng-arpabet strings, including in readalongs. This does not cause any errors that I can see, however.


joanise:
Reduction is coverage % is due to two covered lines being replace by 1 covered line.
New line missed is in test suite, intended to provide better message where there is an error. When the suite passes, that new line isn't supposed to get executed. So coverage should be considered constant for this PR.

littell (approved):

PR #103, opened 2021-04-22
updated CRL/CRJ mappings

deltork:
I have updated the CRL/CRJ mapping to accept the long vowel format.


joanise:
This PR has empty files under `mappings/langs/moe` which cause `g2p update` to fail. I'm not sure you meant to include them.
I think we'll need to create a new PR with only commit `dbe5ac8` on it, but fixed so `g2p udpate` is run from a sandbox without your `moe` files in it. Alternatively, maybe you could include the `moe` mapping too?

joanise:
I need to fix that! From your e-mail, I see that committing `moe` was part of the point of this PR, so something simply went wrong with the merge, because those `moe` files are empty.

deltork:
added some data to the moe lang for now

joanise:
So is the `moe` directory just a placeholder for future stuff? I'd rather delete it for now, until you're ready to put real files in.

joanise:
So, if I understand what you're doing correctly here, you've decided to use `crl_equiv` and `crj_equiv` to map the romanized spelling back to syllabic and funnel them through the existing syllabics -> ipa mappings that already exist. It's a clever solution. My question is, is it accurate to say that  `crl` and `crj` can equally well refer to those languages written in syllabics or romanized? If so, I'll be happy to merge this in, but with the `moe` files deleted rather than containing placeholder stuff.

joanise:
By the way, for future PRs, it would make my life easier if you did the PR from a branch on the main g2p repo, rather than from your fork. From your fork, I can't push additional commits to fix little things, whereas if it was a branch here, I could. For example, I could tidy up the moe stuff myself instead of asking you to do it.

deltork:
> So, if I understand what you're doing correctly here, you've decided to use `crl_equiv` and `crj_equiv` to map the romanized spelling back to syllabic and funnel them through the existing syllabics -> ipa mappings that already exist. It's a clever solution. My question is, is it accurate to say that `crl` and `crj` can equally well refer to those languages written in syllabics or romanized? 


Did that to reduce future maintenance headaches. 
I am tempted to split the crl_equiv file since it is solving two separate issues i.e. keyboarding and roman orthography
Yes for CRL and CRJ the roman and syllabics are now equally covered. The mapping is identical between CRL & CRJ so that users can use either without worry. 

> If so, I'll be happy to merge this in, but with the `moe` files deleted rather than containing placeholder stuff.
I have removed it .. it is something I could use a brainstorming session with the team (maybe next meeting?)

joanise (approved):
Looks good, thanks for tweaking and adding test test cases!

PR #113, opened 2021-05-26
Dev.tli patch

shankhalika-katya:
Converted Tlingit mapping from NFC to NFD


joanise:
I'll take a look at this PR shortly. I know why the autobuild is failling on it, I'll take care of that too.

joanise:
@shankhalika-katya Since you're modifying the tli-ipa to eng-ipa mapping, can you add some test cases in `tests/public/data/tli.csv` with lines like
```
tli,eng-ipa,<tli input>,<eng-ipa output>
tli,eng-arbabet,<tli input>,<eng-arpabet output>
```
which would fail on the code before this patch, but succeed with this patch, thus testing that the change is correct as a whole?
A few whole words on each case would do the trick to validate that it's all working as expected.

joanise (changes requested):
With more testing (see my earlier comment) I believe I will be ready to approve this PR.

joanise:
I just found a bug in the tli mapping.
```
$ g2p convert kawe𝚔̲i𝚐̲ tli eng-arpabet
K AE W EH K ʰIY K
```
That ʰ needs to be mapped out before going to eng-ipa in tli-ipa to eng-ipa, so that we get something correct in arpabet.

I'm guessing rule x->k should come before kʰ -> k, and probably qʰ -> k should come before q -> k. I can see other rules that are probably ordered backwards, e.g., e -> ɛ coming before ẽ -> ẽ will likely prevent the latter from ever being triggered, with NFD normalization anyway. One more, qʼ -> q coming after q -> k might mean the former will never get triggered.

Sorry I'm giving you more work. These mappings are quite tricky, and obviously generate-mapping didn't do a good job here. This is where good unit testing all the way through to eng-arpabet can help guide us in manually fixing the generated mappings. And it shows that careful manual analysis of the generated mappings is sometimes necessary, unfortunately.

shankhalika-katya:
Thanks for pointing this out! I’d just found a similar rule-ordering bug with short and long vowels- I’ll go through the whole mapping line by line now to make sure they’re all ordered properly, and then test through to eng-arpabet.

joanise (approved):
Looks good, thanks for the extra fixes and testing.
I've added a small change: I removed the old `-norm-` token from the filename, which isn't needed here, and has been replaced by `equiv` where it is still relevant.

PR #114, opened 2021-05-27
Dev.optimize test suites

joanise:
For a small drop in code coverage, this PR makes the unit testing suite 20-25s faster.

There were two calls to `g2p doctor` -- one via the CLI, one via the underlying function -- which cost about 25 s together, but did not assert anything useful, so that even a breaking change would probably not have been caught.

So I'm removing these two calls from the test suite: we'll all be happy seeing them run faster, both interactively and in CI.

### DIFF
--- a/g2p/tests/test_cli.py
+++ b/g2p/tests/test_cli.py
@@ -84,6 +84,8 @@ class CliTest(TestCase):
 
         # Disable this test: it's very slow (8s, just by itself) and does not assert
         # anything useful.
+        # Migrated to test_doctor_expensive.py so we can still run it, manually or via
+        # ./run.py all.
         # result = self.runner.invoke(doctor)
         # self.assertEqual(result.exit_code, 0)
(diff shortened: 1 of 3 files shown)


joanise:
A detail about the dropped coverage: it's mostly code under `g2p/tests/` that is no longer run. I only found two lines under `g2p/g2p/` that is no longer covered because of this PR: the line in `cli.py` that logs a message saying "Checking all IPA mappings", and the line `mappings/langs/utils.py` setting `mappings_to_check` to all mappings. (See https://coveralls.io/builds/40084413)

roedoejet (approved):
If it's really not asserting anything meaningful, it's fine to take it out. The other option is to just exclude it from the `dev` TestSuite and have it still accessible through `all` or some other suite. As you wish.

joanise:
Good idea, I've migrated the two expensive tests to `test_doctor_expensive.py` which isn't called from `run.py`, so it won't get run by default, but they'll still be there if we want to run them or improve on them later.

PR #116, opened 2021-06-04
Dev.und: text-unidecode as mapping from und to und-ascii

joanise:
Create mapping und -> und-ascii that maps character by character through text_unidecode, creating the edges along the way and a `TransductionGraph`, like other mappings do.


littell (approved):
To confirm, even though we have mapping metadata for this, we're at no point generating an explicit mapping file for the correspondences; it's all just happening as custom code, right?

joanise:
@littell Correct, it's all in the call to `text_unidecode.unidecode()`. It took me a little while to figure out how to do this correctly, but in the end it boils down to setting `self.mapping = []` in `Mapping` when `type` is `unidecode`, and then having `Transducer` trigger on seeing `type==unidecode`.

PR #118, opened 2021-06-14
chore: g2p update - fixing error rebasing dev.und for PR #116

joanise:
Quick PR fixing error I made while rebasing dev.und to clean up PR #116: I forgot to commit the two `static/*/*.json` files that `g2p update` updated, so that `und-ascii` is missing from these two files in the state as it exists on master.


joanise:
The diff is hard to read, but if you search for `und` in the diff view for this PR, that will highlight `{"source": "und", "target": "und-ipa"}`  being replaced by `{"source": "und", "target": "und-ascii"}` and `{"source": "und-ascii", "target": "und-ipa"}`.

littell (approved):

PR #119, opened 2021-08-03
Dev.tau patch

shankhalika-katya:
tone diacritics are preserved in tau_to_ipa mapping and manually deleted in the tau-ipa_to_eng-ipa mapping
nasality is included in tau_to_ipa mapping
test cases modified to account for the above modifications


joanise:
Build was broken because my GPL test was not working quite right. I'll fix that issue and rebase your branch onto the fixed master, then we'll be able to merge.

joanise (approved):
Thanks for these improvements to the `tau` mapping. I'm glad to have tau-ipa more accurate and the stripping down happening into eng-ipa.

PR #124, opened 2021-10-14
Make g2p generate-mapping accept non-standard mapping names

joanise:
Add optional OUT_LANG to "g2p generate-mapping" so that one can use it even with languages where the mapping goes through an equiv step.

### DIFF
--- a/g2p/cli.py
+++ b/g2p/cli.py
@@ -8,6 +8,7 @@ from flask.cli import FlaskGroup
 from collections import OrderedDict
 from networkx import draw, has_path
 
+from g2p.exceptions import MappingMissing
 from g2p.transducer import CompositeTransducer, Transducer
 from g2p.mappings.create_fallback_mapping import (
     align_to_dummy_fallback,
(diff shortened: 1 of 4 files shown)


g2p/cli.py
```diff
@@ -56,51 +57,95 @@ def cli():
...
+    default=None,
+    #type=click.Choice([x for x in LANGS_NETWORK.nodes if is_ipa(x)]),
+    type=str,
+)
 @click.argument(
     "in_lang",
-    type=click.Choice(
-        [x for x in LANGS_NETWORK.nodes if not is_ipa(x) and not is_xsampa(x)]
-    ),
+    #type=click.Choice([x for x in LANGS_NETWORK.nodes if not is_ipa(x) and not is_xsampa(x)]),
```
roedoejet: If we're replacing this - I think we can remove this instead of uncommenting it.

g2p/cli.py
```diff
@@ -56,51 +57,95 @@ def cli():
...
+            "IN_LANG must be a non-IPA language code with an existing IPA mapping, "
+            f"i.e., one of:\n{', '.join(in_lang_choices)}."
         )
+
+    out_lang_choices = [x for x in LANGS_NETWORK.nodes if is_ipa(x)]
+    if out_lang is None:
+        out_lang = f"{in_lang}-ipa"
+    elif out_lang not in out_lang_choices:
+        raise click.BadParameter(
+            f'Invalid value for OUT_LANG: "{out_lang}".\n'
```
roedoejet, line 112: Nice. Clear, and informative.

g2p/cli.py
joanise: you're right. done

PR #125, opened 2021-10-15
Add standalone mapping configuration flag to g2p convert cli

roedoejet:
This adds the `--config` flag to `g2p convert` to allow for mappings to be added to the language/mappings network without using g2p update. The preferred and quickest way of using g2p is by using g2p update, but for some use cases this might be preferable.


g2p/cli.py
```diff
@@ -95,7 +95,8 @@ def generate_mapping(in_lang, dummy, ipa, list_dummy, out_dir):
         eng_ipa = Mapping(in_lang="eng-ipa", out_lang="eng-arpabet")
         new_mapping = Mapping(in_lang=in_lang, out_lang=f"{in_lang}-ipa")
         click.echo(f"Writing English IPA mapping for {in_lang} to file")
-        create_mapping(new_mapping, eng_ipa, write_to_file=True, out_dir=out_dir)
+        create_mapping(new_mapping, eng_ipa,
+                       write_to_file=True, out_dir=out_dir)
```
joanise, line 99: nitpicking here, but this is not our standard black formatting. I'm not picky about what format we use, but I prefer consistency...

g2p/cli.py
```diff
@@ -172,13 +178,34 @@ def convert(
     # Check valid input
     # Check input != output
     if in_lang == out_lang:
-        raise click.UsageError("Values must be different for 'IN_LANG' and 'OUT_LANG'")
+        raise click.UsageError(
+            "Values must be different for 'IN_LANG' and 'OUT_LANG'")
```
joanise, line 182: Your editor must be configured to auto format to a different format than our standard Black one... This is the second change you make away from our standard Black format, I don't believe you're dong this by hand.

g2p/cli.py
```diff
@@ -172,13 +178,34 @@ def convert(
...
+                    (data['mappings'][index]['in_lang'], data['mappings'][index]['out_lang']))
+                data['mappings'][index] = load_mapping_from_path(config, index)
+        else:
+            data = load_mapping_from_path(config)
+        for pair in mappings_legal_pairs:
+            if pair[0] in LANGS_NETWORK.nodes:
+                LOGGER.warn(
+                    f"A mapping with the name '{pair[0]}' is already defined in g2p. Your local mapping with the same name might not function properly.")
+        LANGS_NETWORK.add_edges_from(mappings_legal_pairs)
+        MAPPINGS_AVAILABLE.extend(data['mappings'])
```
joanise, line 200: So, if I understand this code correctly, `--config` adds to the existing network, rather than wholesale replacing the existing network. That seems much more convenient, so I agree with the idea, I had just understood it the other way around, that only what's in the given config would be usage.

g2p/cli.py
joanise: In fact, there are many more changes lower in this file that keep the code identical but reformat away from our Black settings. If I come over and apply black later, I'll end up undoing a bunch of these changes, it's better if you fix your editor and patch this PR now instead.

g2p/tests/test_cli.py
```diff
@@ -166,9 +171,19 @@ def test_convert_option_t(self):
...
 
+    def test_local_config(self):
+        config_path = os.path.join(PUBLIC_DIR, 'mappings', 'test.yaml')
+        result = self.runner.invoke(
+            convert, f"bbbb local-config-in local-config-out --config {config_path}")
+        self.assertIn("aaaa", result.stdout)
+        result = self.runner.invoke(
+            convert, f"b local-config-in eng-ipa --config {config_path}")
+        self.assertIn("ɑ", result.stdout)
+
```
joanise, line 186: You cannot use literal `/` as path joiner, you have to use os.path.join. On my Windows machine:
```
$ ./test_cli.py
INFO - Running 355 g2p convert test cases found in public/data
.......FFE.
======================================================================
ERROR: test_scan_str_case (__main__.CliTest)
----------------------------------------------------------------------
  File "./test_cli.py", line 140, in test_scan_str_case
    returned_set = re.search('{(.*)}', result.stdout).group(1)
AttributeError: 'NoneType' object has no attribute 'group'

======================================================================
FAIL: test_local_config (__main__.CliTest)
----------------------------------------------------------------------
  File "./test_cli.py", line 182, in test_local_config
    self.assertIn("aaaa", result.stdout)
AssertionError: 'aaaa' not found in "Usage: convert [OPTIONS] INPUT_TEXT IN_LANG OUT_LANG\nTry 'convert -h' for help.
\n\nError: Invalid value for '--config': File 'c:usersjoaniserasg2pg2ptestspublicmappingstest.yaml' does not exist.\n
"

======================================================================
FAIL: test_scan_fra_simple (__main__.CliTest)
----------------------------------------------------------------------
  File "./test_cli.py", line 128, in test_scan_fra_simple
    self.assertEqual(result.exit_code, 0)
AssertionError: 2 != 0

----------------------------------------------------------------------
Ran 11 tests in 10.011s

FAILED (failures=2, errors=1)
```

joanise: Hum, it fixes one instance, which is still showing as an error here, but 6684cbb79 replaces a bunch of `os.path.join()` by f-strings with hard-coded slashes, in particular in `not_test_scan_fra()`, `test_scan_fra_simple()` and `test_scan_str_case()`. I don't get why you would have made these changes, actually, maybe they're other auto edits?

But I wonder, you're using os.path.join to construct the file name, then building a command line from it. I normally don't build the command line as f-strings in my unit testing, but rather as an array of argument. This might make a difference on Windows it a way it might not be relevant on *nix.

joanise (approved):
I'm happy with this PR now. Running the test suites no longer has unintended side effects.
Nice feature, as I see while reading the blog post.

PR #135, opened 2021-11-15
Two small patches

joanise:
Improving the fra g2p mapping, and the `g2p generate-mapping -h` message.


roedoejet (approved):
looks good @joanise

PR #136, opened 2021-11-17
refactor: make_g2p should raise more meaningful exceptions on caller errors

joanise:
Change the exceptions raised by `make_g2p()` to NoPath and InvalidLanguageCode instead of NetworkXNoPath and FileNotFoundError.

Rationale:
 - the two exceptions are part of the g2p module and can be imported like `from g2p import make_g2p, NoPath, InvalidLanguageCode`
 - they're more intuitive (I think)
 - and I added documentation on `make_g2p()` with the `Raises:` field so these exceptions are documented as part of the API.

### DIFF
--- a/g2p/__init__.py
+++ b/g2p/__init__.py
@@ -9,6 +9,7 @@ import io
 from networkx import shortest_path
 from networkx.exception import NetworkXNoPath
 
+from g2p.exceptions import InvalidLanguageCode, NoPath
 from g2p.mappings import Mapping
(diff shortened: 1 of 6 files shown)


joanise:
@roedoejet This should be a quick review now. I thought you had already reviewed it, actually...

roedoejet (approved):
Huh, I thought I reviewed this too. I guess I just messaged on slack about it?

joanise:
Yeah, we did discuss it on Slack. Thanks for approving it.

PR #138, opened 2021-11-18
fix(deployment): replace eventlet with gevent

roedoejet:
I'm getting sick of waiting for eventlet, so we can replace with gevent (for now)


joanise:
Just starting to review, but it doesn't have the dnspython and eventlet versions I thought you were going to use.

requirements.txt
```diff
@@ -1,15 +1,18 @@
...
-eventlet==0.30.2
+dnspython>=2.1.0
+eventlet>=0.33.0
 flask-cors>=3.0.9
 Flask>=2.0.0
 flask_socketio==4.3.2
 flask-talisman>=0.7.0
 flask-restful>=0.3.9
-gunicorn
+gunicorn @ git+https://github.com/benoitc/gunicorn.git@ff58e0c
```
joanise, line 11: ugh, I just did the installation, and there's an unfortunate side effect here that I didn't expect: we're making `git` a dependency that has to be installed before one can do even a vanilla `pip install readalongs` or `pip install g2p`. Contributors will have it, but we shouldn't require it for users.

It's pretty cool how pip will actually clone the repo and build from it. I had no idea that was possible. I'd seen this syntax before, but I didn't clue in to what it actually does.

One solution I can imagine: go back to just `gunicorn` here, and add this command to Dockerfile and to the dot file that controls heroku: 
`pip install gunicorn@git+https://github.com/benoitc/gunicorn.git@ff58e0c`

It's not a great solution, because we'd want something cleaner that also works for interactively launching the gunicorn server.

A second solution, which you suggested yesterday: add a `requirements.deploy.txt` with the dependencies that are only required for running with gunicorn. And `pip install -r` that file in Dockerfile, .heroku, and mention it wherever in the doc we say how to launch with gunicorn.

Actually, I have a strong preference for this last solution, because it would reduce the dependency bloat we are getting when doing `pip install readalongs`. In an ideal world, that command should only install what is actually needed to run `readalongs <cmd>`, as well as `python run_studio.py`, with as few extras as possible.

roedoejet: so Heroku is more strict than I thought. sounds like it basically requires that the file be called requirements.txt

Seems like people create a requirements folder and then have the root requirements.txt file mirror some `requirements/requirements.prod.txt` file, and then change installation instructions/setuptools etc to point to `requirements/requirements.dev.txt` instead. I don't love that, but I suppose maybe it's better than introducing a dependency on git or waiting until gunicorn decides to push a release?

setup.py
```diff
@@ -1,43 +1,42 @@
...
+with open(
+    path.join(this_directory, "requirements", "requirements.txt"), encoding="utf8"
+) as f:
     REQS = f.read().splitlines()
 
 setup(
-    name='g2p',
-    python_requires='>=3.6',
+    name="g2p",
+    python_requires=">=3.6",
```
joanise, line 28: Is >= 3.6 still true? I thought we decided >=3.7? Or was that only for Studio?

joanise (approved):
Looks good! thanks!

setup.py
roedoejet: g2p still works for 3.6!

joanise: Ah, OK. Then I guess we should keep it at 3.6 for sure. I suppose there's not motivation to move on unless we want to use 3.7 features.

PR #139, opened 2021-11-25
fix(mappings): reverse length sort abbreviations to prevent substring errors

### LINKED ISSUE #133: Abbreviation names cannot be substrings
If I create a mapping with abbreviations, and use names that are substrings of each other, it breaks the interpretation.

To reproduce, modify `mappings/langs/fra/fra_abbs.csv` and rename `AOU_VOW` to `AOU_VOWEL` and `IE_VOW` to `EI_VOWEL`. Make the corresponding changes in `fra_to_ipa.csv`, run `g2p update`, then `tests/test_langs.py` now fails.

Running `g2p convert --debugger mangeons fra fra-ipa` shows that `AOU_VOWEL` was expanded to `AOU_{value of VOWEL}` instead of `{value of AOU_VOWEL}`.

roedoejet:
fixes #133

### DIFF
--- a/g2p/mappings/__init__.py
+++ b/g2p/mappings/__init__.py
@@ -114,7 +114,7 @@ class Mapping:
         self.processed = False
         if isinstance(abbreviations, defaultdict) or not abbreviations:
             self.abbreviations = abbreviations
-        elif abbreviations:
+        else:
             self.abbreviations = load_abbreviations_from_file(abbreviations)
         # Handle user-supplied list
         if isinstance(mapping, list):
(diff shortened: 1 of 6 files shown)


joanise (approved):
Looks good. Looking at the diff via https://github.com/roedoejet/g2p/pull/139/commits/9e25a2b825d048dd9a5377ce24e0d40111a805f9
made it clear what you changed, and it all makes sense. Much simpler than I expected.

roedoejet:
thanks @joanise !

PR #145, opened 2022-02-28
fix(api): add index and debugger flags to documentation, add localhos…

roedoejet:
In looking at the API documentation I realized the index and debugger query args weren't part of the swagger documentation. I've added those, and fixed some tests.


joanise:
With the black reformatting, I don't see what has actually changed... :(

joanise (approved):
Looks good.

Issue #147, opened 2022-03-09
Change default Unicode normalization to NFC


roedoejet:

roedoejet:
Currently default is NFD based on stale downstream requirement from ReadAlongs - but NFC generally renders better and should be used by default.

joanise:
Just a thought: we could go through all the existing mappings that don't explicitly declare their norm-form, and explicitly declare then NFD if that's actually needed. Then, changing the default should not have unintended consequences.

Oh, boy, just checked, and there are 55 mappings that don't declare their norm_form. This will be a much bigger job than I imagined...

joanise:
With the compose-norm PR I just did, this might now be a safe things to do.

PR #148, opened 2022-03-10
NFC as default for from/to gen'd mappings; NFD testing for g2p scan

joanise:
Quick PR:
 - make NFC the default for new generated mappings - this does not fix the larger issue (#147), it only applies NFC to new mappings generated in the from/to mode.
 - expanding the unit test for `g2p scan` to cover both NFC and NFD - that's stuff I prepared months ago but forgot about. I just noticed it this morning cleaning up old sandboxes.


roedoejet (approved):
Looks good - thanks @joanise 

This fixes one part of https://github.com/roedoejet/g2p/issues/147

PR #149, opened 2022-03-29
Dockerfile optimizations

joanise:
Bump the OS to the current stable, bullseye
Optimize building the docker image
Lock werkzeug to 2.0.3, since our other dependency versions are not compatible with the latest, 2.1.0


roedoejet (approved):
thanks @joanise

PR #150, opened 2022-04-12
fix: make Mohawk tokenizer recognize colon as a letter

joanise:
The tokenizer was not recognizing the colon as a letter for Mohawk, the way it should have.

This solution is somewhat hacky: instead of fixing the tokenizer code, I'm just adding a dummy map from `:` to `:` in equiv. This may be a bit lazy, but it's effective and reliable.


joanise:
The problem was that in the moh g2p mapping, : is always in specific contexts, never by itself. This quick fix brings in a potentially problematic change: if : occurs in a Mohawk file as a colon, or after a letter where it's not supposed to be, then it'll end up staying a colon in the ARPABET output, and an accompanying error message.

So the question is, before merging this PR: is colon ever used as a punctuation in Mohawk text? If so, we need a better solution,

PR #161, opened 2022-06-07
ci: run the CI tests on pull requests too

joanise:


dhdaines (approved):
This looks correct to me, at least, it's valid YAML and seems to be in line with https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions

PR #162, opened 2022-06-07
fix: lock click==8.0.4 since we support Python 3.6

joanise:
The click change log https://python.libhunt.com/click-changelog says
that release v8.1.0 drops support for Python 3.6, but we're still
officially supporting Python 3.6 for g2p, so we need to lock the click
version accordingly.

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -1,5 +1,5 @@
 openpyxl
-click
+click==8.0.4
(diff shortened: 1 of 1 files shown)


dhdaines (commented):
seems good to me, and the tests pass with python 3.6

dhdaines (approved):

PR #163, opened 2022-06-09
feat: g2p show-mappings

joanise:
When debugging g2p, I often find I'd like to know what's actually in the cached mappings at a given commit or in a given sandbox. `g2p show-mappings` does just that: dump to screen the stuff that `g2p update` last saved.

Other change: when `context_before` and `context_after` are actually empty, don't print to with `g2p show-mappings` and don't include them in the files generated by `g2p generate-mapping`.

### DIFF
--- a/g2p/cli.py
+++ b/g2p/cli.py
@@ -1,6 +1,8 @@
+import json
 import os
 import pprint
 import re
+import sys
 
 import click
 import yaml
@@ -11,7 +13,7 @@ from g2p import make_g2p
 from g2p._version import VERSION
 from g2p.api import update_docs
 from g2p.app import APP, network_to_echart
-from g2p.exceptions import MappingMissing
+from g2p.exceptions import InvalidLanguageCode, MappingMissing, NoPath
 from g2p.log import LOGGER
 from g2p.mappings import Mapping
 from g2p.mappings.create_fallback_mapping import (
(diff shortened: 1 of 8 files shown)


g2p/cli.py
```diff
@@ -677,3 +679,59 @@ def scan(lang, path):
...
+            Mapping(in_lang=m["in_lang"], out_lang=m["out_lang"])
+            for m in MAPPINGS_AVAILABLE
+            if m["in_lang"] == lang1 or m["out_lang"] == lang1
+        ]
+        if not mappings:
+            raise click.BadParameter(
+                f'No mapping found to or from "{lang1}".', param_hint="lang1"
+            )
+
+    else:
```
roedoejet, line 708: I think this is a bit cryptic - if a user types `g2p show-mappings` without the --all flag, it will just say "nothing to do!". This message should either tell the user to use the --all flag or (better yet) remove the all flag and print the total mappings in the else case.

roedoejet (commented):
@joanise - I went ahead and added the changes I would like to see, but if you don't agree with these changes then of course let's have a chat on slack. I also removed the old `g2p generate-mapping-network` command which was intended to do the same thing sort of.

roedoejet:
> @joanise - I went ahead and added the changes I would like to see, but if you don't agree with these changes then of course let's have a chat on slack. I also removed the old `g2p generate-mapping-network` command which was intended to do the same thing sort of.

Basically, I made it so you don't need the `--all` flag, and I also made it so that your output is returned with the `--verbose` flag but a much terser output of just the input/output is returned by default.

joanise:
Looks good. Makes sense to have "all" be the default instead of requiring a switch. This will make my `g2p doctor --list-all` and `--list-ipa` redundant, too, although I would not rush to delete them yet.

joanise:
@roedoejet I make a small additional change. I like your suggestions. Ready to re-view and merge as far as I'm concerned.

roedoejet (approved):
Nice addition - thanks @joanise !

PR #164, opened 2022-06-09
ci: publish with build instead of calling setup.py directly

joanise:
According to https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html
calling `python setup.py` has been deprecated for a while, and should be
avoided. The recommended replacement is

    python -b build --sdist --wheel

which still uses the setup.py file, but in the way that is now considered
correct.  This commits introduces that method. The build command was tested in
ReadAlongs/Studio branch dev.test-build, with the results visible here:
https://github.com/ReadAlongs/Studio/actions?query=branch%3Adev.test-build

Other changes:
 - have setup.py declare platform and classifiers

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -17,13 +17,13 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


dhdaines (approved):
This is great to know!

PR #166, opened 2022-06-14
feat: include NFC/D normalization in g2p graph

### LINKED ISSUE #158: Composed indices don't make sense when normalization changes along the path
On branch `dev.compose-bug`, I have an NFC mapping with `é,ò` chained with an NFD mapping with `ò,ù`.

When I compose the two, the composed indices don't make sense:

```
$ g2p convert --config public/mappings/compose.yaml é c1 c3 -d -e
[   'ù',
    [[['é', 'ò']], [['o', 'u'], ['̀`' '̀`']]],
    [[(0, 0)], [(0, 0), (1, 1)]],
...
```
The problem is with the `(1,1)`: it's coming out of nowhere, since the output of the first step only had one element.

We would need an explicit NFD step with its own indices to make this make sense, to produce output like:
```
[   'ù',
    [[['é', 'ò']], [['ò', 'o'], ['o', '`']], [['o', 'u'], ['̀`' '̀`']]],
    [[(0, 0)], [(0, 0), (0, 1)], [(0, 0), (1, 1)]],
...
```
Where I've manually inserted indices to map the one-character NFC "ò" to the two-character NFD "ò".

This brings up a whole challenging set of design question, though: it would mean that suddenly, the indices returned by a 2 step mapping could have more than two sets of indices.

Maybe instead of doing this, we would want to pre-compose the results of the normalization into the indices that get printed. I.e., since the composition of `[(0, 0), (0, 1)]` and `[(0, 0), (1, 1)]` is `[(0, 0), (0, 1)]`, we would just output the latter. In that case, the output of my convert command would be:
```
[   'ù',
    [[['é', 'ò']], [['o', 'u'], ['̀`' '̀`']]],
    [[(0, 0)], [(0, 0), (0, 1)]],
...
```

Note: I'm cheating above for legibility's sake. I replaced the accent grave by backticks, so we can see them, but the real output looks like `['̀', '̀']` since the accent grave is a combining character.

joanise:
Previously, the graph generated by g2p convert was from the string
passed in after normalization. And if the path had several steps, the
graph between these steps might have been incoherent if the steps did
not all use the same normalization (see #158).

This PR modifies the g2p transducer class to generate the graph from
the input to convert to the output, by considering the NFC/D
normalization like a transduction step in itself, and composing the
indices it generates with the ones generated by the mapping itself.

FIXES #158


joanise:
For comparison, when I convert `abcéàçù` from `c1` to `c3` with the mappings from `public/data/compose.yaml` added to the cache.

The mapping from c1 to c2 is NFC, while the mapping from c2 to c3 is NFD, so the initial input is NFC normalized, then the output of c1->c2 is NFD normalized before the c2->c3 mapping is applied.

With this setup, the graphs illustrate the changes this PR brings quite clearly.

Using the code in the master branch:
image

Using the code in this PR:
image

 - `abc` gets a weird graph through to `de` because my c1->c2->c3 rules deliberately do that.
 - `é` goes to `ò` and then `u` plus the accent as a separate character because of NFD.
 - I included the extra composed characters `àçù` to show how the NFD conversion ends up generating explicit conversions in this PR but nonsense in master.

roedoejet (approved):
great feature @joanise - thanks! very concise and clean solution.

PR #167, opened 2022-06-16
fix: generated mappings should prevent feedback and apply-longest-first

joanise:
Long-standing fix I've been intending to do: `g2p generate-mapping --ipa` should also generate mappings with prevent feeding and apply longest first, to avoid bugs in the generated mappings like Shanks and Sabrina discovered in their work.


roedoejet (approved):
makes sense. thanks @joanise

PR #168, opened 2022-06-16
Fix win mapping with right length marker and avoid doubling it

### LINKED ISSUE #100: win does not handle : correctly
1. The `win` g2p outputs ascii : instead of `\u02D0`, which is the correct IPA symbol for length markers.

2. The win-ipa to eng-ipa mapping sometimes doubles them up. All the lengthened vowels needs to be reviewed.

e.g.:
```
$ g2p convert "ō" win eng-arpabet
OW ː
$ g2p convert "ee" win eng-arpabet
EY ː
```

3. There is no `public/data/win.*` test suite to validate that g2p works correctly on this language.

joanise:
The length marker in IPA is ː (\u02D0), not : (ascii colon)

win-ipa to eng-ipa had a feeding problem, with `"eː"->"eː"` feeding
`"e"->"eː"` and thus generating `"eːː"` in the eng-ipa output, and `"EY ː"` in
the eng-arpabet output.

Unit tests added for individual characters, but I have no words to test
properly, so I'm doing the best I can with what I have...

FIXES #100


roedoejet (approved):
Yep, good easy fix. Thanks for catching this.

PR #169, opened 2022-06-16
Blackify and isort the whole g2p code base

joanise:


joanise:
This PR touches most python source files, but it does not introduce conflicts with #167 or #168, so my three open PRs can be reviewed and merged in any order.

roedoejet:
We should probably add these things as pre-commits then eh (like in https://github.com/ReadAlongs/Studio/blob/master/.pre-commit-config.yaml along with a .flake8 file) and then update the installation instructions with guidance for advanced users. I don't think we should enforce installing pre-commit hooks for everybody, it's an extra layer that goes against our goals of making it accessible, and it's easy enough to run black and isort on someone else's PR, but we should have the option, and a little language about how to install them. What do you think @joanise ?

joanise:
I agree, it should be optional for this repo. I was thinking I wouldn't automate any of it, but you're right, it's a good idea to do the same automation as for Studio, but just skip the CI check.

roedoejet:
> @roedoejet I'm starting to add the pre-commit hooks and the Contributing.md file, mimicking and adjusting what we have in Studio. (Let me know if you were already doing that, I don't want to duplicate work.)

Just getting back from lunch and hadn't done it yet. Thanks!

PR #170, opened 2022-06-17
fix: make und work in g2p studio

### LINKED ISSUE #165: und does not work correctly in g2p Studio
In CLI:

```
$ g2p convert éçà und eng-arpabet
EY CH AA
```

In g2p Studio, select:
 - Input language = `und`
 - Output language = `eng-arpabet`
 - Input text = `éçà`

The output shown is `éCH ̧à` instead of `EY CH AA`

Using `eca` as input, the CLI correctly outputs the same `EY CC AA`, but the g2p Studio outputs the also incorrect `eCH a`.

joanise:
The reason und did not work (see issue #165) is that class Transducer
relies on having type == "unidecode" to trigger the und unidecode
processing.

Solution: add an input field for "type", which defaults to "mapping",
but is set to "unidecode" when we load the und->und-ascii mapping from
cache.

Now, that input is really specific, only applicable to the und -> und-ascii
mapping, so hide it and don't let the user play with it.

FIXES #165

### DIFF
--- a/g2p/static/custom.js
+++ b/g2p/static/custom.js
@@ -59,6 +59,7 @@ function createSettings(index, data) {
     let out_delimiter = '';
     let prevent_feeding = '';
     let norm_form = 'NFC';
+    let type = 'mapping';
     if (index === 0) {
         active = 'active'
     }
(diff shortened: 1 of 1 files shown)


roedoejet (approved):
yep - this solves it! was gonna look into this this morning but you beat me to it - you're on fire @joanise !

PR #171, opened 2022-06-17
pre-commit hooks and commitlint

joanise:
Activate and document pre-commit hooks and commitlint


package.json
```diff
@@ -0,0 +1,35 @@
+{
```
roedoejet, line 1: It looks like there is a python package now for conventional commits - so maybe we don't need npm/husky here anymore? https://pypi.org/project/conventional-commit/

I don't really mind, but adding it to pre-commit would reduce this dependency

joanise: Cool, that's worth looking into. It would be nice to remove the node dependency from both Studio and g2p, since it's only for commitlint.

joanise: Hmm, no, it's an interactive interface to compose commit messages, not a linter. Searching for alternatives, maybe this is a candidate: https://jorisroovers.com/gitlint/

joanise:
@roedoejet I found and configured gitlint. It's quite nice, really.
1) Only python dependencies, so `pip install -r requirements/requirements.dev.txt` installs it.
2) after validating your commit, if there's an error it prompts you to edit, cancel or commit anyways, which is a small difference but much nicer UX than commitlint.

roedoejet:
looks good @joanise - thanks!

PR #177, opened 2022-07-09
Basic mapping for Finnish

dhdaines:

- [ x ] Mapping files are added in `g2p/mappings/langs`
- [ x ] Mapping is either added to an existing folder or a new folder has been added
- [ x ] Language folder and files use appropriate [ISO 639-3 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-3_codes)
- [ x ] Config.yaml file includes all author names, and settings necessary
- [ x ] Please add some test data in `g2p/tests/public/data`. The added file should be a csv/tsv/psv file and each row should have the format `[input_mapping_code,output_mapping_code,input_string,output_string]`
- [ x ] As the last step, G2P has been updated by running `g2p update` locally and committing the change
- [ x ] You agree to license your contribution under the same license as this project (see [LICENSE](https://github.com/roedoejet/g2p/blob/master/LICENSE) file).

* **Other information**:


g2p/mappings/langs/fin/fin_to_ipa.csv
```diff
@@ -0,0 +1,50 @@
...
+z,t͡s,,
+å,o,,
+ää,æː,,
+ä,æ,,
+öö,øː,,
+ö,ø,,
+n,ŋ,,k
+n,ŋ,,g
+n,m,,p
+n,m,,b
```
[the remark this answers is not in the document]
joanise: Without prevent-feeding, some of your vowel rules scare me a bit, but I analyzed them all, and it doesn't look like any vowel rule will actually destroy the output of other vowel rules, so it's probably going to work OK.

g2p/mappings/langs/generated/config.yaml
```diff
@@ -404,15 +404,15 @@ mappings:
...
-    mapping: moe-ipa_to_eng-ipa.json
+    in_lang: fin-ipa
+    language_name: fin-ipa
+    mapping: fin-ipa_to_eng-ipa.json
     norm_form: NFC
     out_lang: eng-ipa
     prevent_feeding: true
     reverse: false
-    rule_ordering: apply-longest-first
+    rule_ordering: as-written
```
joanise, line 418: This should have been added at the end of the file, but accidentally it replaced the `moe-ipa` block instead. We'll need to restore the original `moe-ipa` block that was in `master` and add this `fin-ipa` block after it, rather than have the `fin-ipa` block replace the `moe-ipa` block.

g2p/tests/public/data/fin.psv
```diff
@@ -0,0 +1,6 @@
+fin|fin-ipa|nähdä|næhdæ
+fin|fin-ipa|polkupyörä|polkupyøræ
+fin|fin-ipa|haluttaa|haluttɑː
+fin|fin-ipa|tulevaisuus|tuleʋɑjsuːs
+fin|fin-ipa|pitämättömyyksien|pitæmættømyːksjen
+fin|fin-ipa|mihinkään|mihiŋkæːn
```
joanise, line 6: I would add some lines to unit test fin -> eng-ipa and/or fin -> eng-arpabet too. I find the generated mappings need a bit of testing too, I've come across bugs when testing the chain.

joanise (changes requested):
See change requests in the interspersed comments.

g2p/mappings/langs/generated/config.yaml
dhdaines: I'm confused about how to write the `config.yaml` - in Aidan's instructions the `<<: &shared` and `<<: *shared` are not there, and I also think they are a bit confusing for the user, but if I don't put them in, then `g2p generate-mapping fin --ipa` fails:

```
(py3-readalongs) dhd@minipc:~/work/g2p$ g2p generate-mapping fin --ipa
  File "/home/dhd/work/py3-readalongs/bin/g2p", line 33, in <module>
    sys.exit(load_entry_point('g2p', 'console_scripts', 'g2p')())
  File "/home/dhd/work/py3-readalongs/bin/g2p", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib/python3.8/importlib/metadata.py", line 77, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 961, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 848, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/dhd/work/g2p/g2p/__init__.py", line 26, in <module>
    from g2p.mappings import Mapping
  File "/home/dhd/work/g2p/g2p/mappings/__init__.py", line 23, in <module>
    from g2p.mappings.langs import LANGS, MAPPINGS_AVAILABLE
  File "/home/dhd/work/g2p/g2p/mappings/langs/__init__.py", line 62, in <module>
    LANGS_AVAILABLE = [
  File "/home/dhd/work/g2p/g2p/mappings/langs/__init__.py", line 63, in <listcomp>
    {k: v["language_name"]}
```

If there are strange side effects of the ordering of keys in `config.yaml` then we should probably fix that in the code that reads the file, if possible ...

joanise: I don't understand why you had to write this block in the first place. `g2p generate-mapping --ipa fin` should have inserted it here for you, didn't it?

as for the `shared` business, that's not what's required: you have to have a `language_name` field, whether in the shared block or directly.

dhdaines: No, I was talking about `g2p/mappings/langs/fin/config.yaml` - that isn't auto-generated, is it?  Sorry, I may have put my question in the wrong place, though I think the generated yaml is ultimately generated from that one, right?

dhdaines: And then the other question is ... *why* did it replace moe-ipa?  I didn't ask it to do that... is `g2p generate-mapping` buggy?

joanise: Right, so no, `fin/config.yaml` is not autogenerated, you did have to do that by hand.

But I have to agree, the example config you show with `- language_name: Finnish` ought to work. Is this what #179 now allows?

joanise: As to why `g2p generate-mapping` replaced `moe-ipa`, I have no idea. That does sound like a but. Any chance you'd be able to pinpoint a specific commit where that behaviour happens? It would be worth creating a branch like `bug.g2p-gen-map` and an issue referring to it, so we can try to fix it at some point.

dhdaines:
This should be okay to merge now

g2p/mappings/langs/fin/fin_to_ipa.csv
```diff
@@ -0,0 +1,50 @@
...
+ii,iː,,
+ie,je,,
+i,i,,
+j,j,,
+k,k,,
+l,l,,
+m,m,,
+n,n,,
+oo,oː,,
+ou,oʊ,,
```
roedoejet, line 24: wow - surprisingly straightforward! I won't look into this, I'll just take your word it's this transparent!

roedoejet (approved):
Thanks @dhdaines !

g2p/mappings/langs/generated/config.yaml
roedoejet: I'm also curious if this can be replicated...

g2p/mappings/langs/fin/fin_to_ipa.csv
dhdaines: A native speaker could certainly clarify some of the vowels a bit, and there are some nuances of gemination / lengthening in some contexts, but yes, it's basically IPA :)

Note that this is the literary language, what people actually speak is ... different

PR #178, opened 2022-07-11
Dev.fix single mappings

dhdaines:
Various problems with defining mappings in the way that they are defined, for instance, on Aidan's blog: https://blog.mothertongues.org/g2p-advanced-mappings/

It should also be possible just to define a single mapping in a file, not a list, which works here as well.

Test coverage is inadequate, I'll get to that in a bit...


dhdaines:
Oh fuck it, I fucked up with Git again.  Closing this to do it right from scratch because I have no goddamn clue how to rebase things "comme du monde"

PR #179, opened 2022-07-11
Make `g2p update` robust to errors and recoverable

dhdaines:
Okay, done properly, the only way I know how (nuke the old branch and make a new one from scratch).

I'm not really sure how to test this properly though as it would require to create a separate test tree of languages for the various cases.  Thoughts?


dhdaines:
> Question: when you run `g2p update`, do mappings added the way you now support get cached properly? I.e., are they available to `g2p convert` afterwards? That might be more challenging to unit test, but at least make sure you manually test it.

Yes, that is precisely what I need to unit test (see #181).  I will need to add an option to put the `langs.pkl` somewhere other than the default location, I think.

joanise:
We've decided to use `black` and `isort` for both `Studio` and `g2p`. If you install the pre-commit hooks (see [Contributing.md](https://github.com/roedoejet/g2p/blob/master/Contributing.md)) it'll happen automatically when you commit.

dhdaines:
In part because the files that I changed were clearly *not* formatted with `black` before I changed them, so I wasn't aware that we were requiring it, and also I didn't want to confuse the diff with formatting changes.  Should I reformat in this PR or should I do a separate one for it?

joanise:
No, no need to create a new PR. Just add one commit to the PR with the four files reformatted.

I'm surprised the files were not blacked before you started, though, because I blackified everything with PR #169 back in mid June. I suppose I might have missed some stuff...

joanise (approved):
This PR looks great now, I think it's ready to merge.

PR #182, opened 2022-07-18
Accept input from G2P Studio

dhdaines:
Fix read_from_csv to remove UTF-8 BOM characters that are introduced by G2P Studio, so that we can actually use G2P Studio to create mappings :)

(this is not an issue for JSON, etc, since they are structured data formats which aren't affected by a useless character at the beginning of the file)

Also test the case where someone may have concatenated CSV files together with stray BOMs in them, except that this will create other problems, since they may not end in newlines, but it's good to be robust in any case.


joanise:
Oh, here too `test_mappings.py` should be blackified.

joanise:
@dhdaines I remember you suggesting the use of `encoding="utf-8-sig"` instead of this solution, do you remember why you ended up using `translate(remove_bom)` instead?

Del just had a use case face plant with a BOM in his config.json file (yeah! - readalongs says: Error: Invalid value: Config file at data/sample-config-bom.json is not in valid JSON format.) and I'm thinking of simply opening the config file with `utf-8-sig`, but I was wondering if there might be cases that won't catch.

joanise:
Hum, I think I remember now: `utf-8-sig` ignores the file-initial BOM, but not mid-file ones, while your solution ignores them everywhere. I see you have a unit test case for that, even.

dhdaines:
Aha, thanks for refreshing my memory on that...

For reading a file, `utf-8-sig` is appropriate - we decided (I think you also discussed this with me!) that in the event that somebody concatenated files with BOMs on them, it was better to remove them everywhere

dhdaines:
Note however, that if you try to concatenate CSV files, you will cause a lot of trouble beyond just random BOMs, as generally Excel produces CSV files without a final CRLF

joanise:
Right, yes, I remember having that discussion now. Thanks for the refresher about the problems concatenating CSV files, I had not really grasped the meaning of the comment about that right in the initial PR description above, but now it all makes sense.

So, this PR stands as is, and I'll use `utf-8-sig` elsewhere when we're reading files that might have the BOM.

PR #183, opened 2022-07-19
Add CI testing on Windows

joanise:
Add a second test job in CI to run the test suite on Windows too, since we intend to support Windows.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,9 +6,9 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Set up Python
-        uses: actions/setup-python@v1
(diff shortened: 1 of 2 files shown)


dhdaines:
Oh wow!  You can do that?!  This will solve many problems for me...

joanise:
Yeah, I just decided to search for this. GitHub belongs to MS now, after all, so I thought it couldn't not be supported...

dhdaines (approved):

joanise:
`windows-latest` is the latest version of Windows Server, (2022 right now), so it's not the same machine normal users would use, but it's still in the Windows ecosystem and I can already see that it's catching those errors due to not declaring the encoding when you open a file.

joanise:
@dhdaines All the available OS choices are documented here: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

roedoejet (approved):

PR #184, opened 2022-07-19
Do full matrix CI on PRs to release

joanise:
Before making a release, let's run the test suite on Windows, MacOS and Ubuntu, with Python versions 3.6 through 3.10.

See https://github.com/roedoejet/g2p/actions/runs/2700305653 for what the action looks like when it fully runs.


roedoejet (approved):
Great! This will let us release a bit more confidently. Thanks @joanise !

PR #185, opened 2022-07-20
Tweak Finnish mappings slightly

dhdaines:
Improve the diphthong mappings (at least, according to my ears, a real linguist or native speaker can surely fix them further).

Question: the generated English IPA mappings are slightly dubious (iː mapping to eː in particular).  Can we edit them?


joanise:
@dhdaines Yes, you can manually edit the generated mapping, for sure. When I do that, I like to add a line under authors saying I did so, in `generated/config.yml`. See the Tlingit IPA to English IPA generated mapping for a good example. But this is not required. In general, Once you've committed a generated mapping, with or without edits, no one will touch it unless there's a problem to fix. It's a good idea to add tests that exercise your manual changes in `tests/public/data`.

joanise (approved):
Looks good to me. Approving this now, but you can do your edits to the generated mappings before merging.

PR #186, opened 2022-07-20
Dev.style

joanise:
Misc style improvements from black, isort and flake8

We still have more flake8 and mypy warnings, but this PR makes some progress anyway.

Also, check the python version (>=3.6) explicitly, like we do in Studio, and lock mypy to <=0.971 since the next one won't support Python 3.6.


roedoejet (approved):
Nice cleanup PR, thanks @joanise !

PR #190, opened 2022-09-08
fix: rules with alternations should tokenize correctly

joanise:
Previously:
  `"in": "é|e'"`
would be considered as a whole, like that, by the tokenizer. But the correct interpretation is to add both `é` and `e'` to the language's input inventory.

This fixes tokenization of Mi'kmaw when `í` is written `i'` instead of `í` or `iꞌ` (i.e., `i\uA78C`). All three forms should be considered equivalent, and should get both tokenized and g2p'd the same way.


joanise:
@roedoejet Once again pushing to codecov is failing. The test suites pass for this PR, but the push to codecov fails. It seems to me last time that just fixed itself. Anyway, I did the coverage analysis on my own machine and I'm satisfied with the results (new stuff that's not covered is to report more explicit errors if a user sets up data incorrectly in `tests/public/data` and if some of the test cases fail. That code will never get covered when the tests pass, by definition.

joanise:
I've added some refactorings to this PR. They might seem unrelated, but they helped me test the changes I made to fix Mi'kmaw tokenization, so it makes sense to me to keep them in. I could do some squashing of the three refactor commits, since it took me three tries to get it right, but it doesn't seem necessary to me.

joanise:
OK, I just tidied up the history squashing that refactoring work into just one sensible commit. Did a forced push to update the PR accordingly.

g2p/mappings/tokenizer.py
```diff
@@ -67,7 +67,14 @@ def __init__(self, mapping: Mapping):
     def _build_regex(self):
         if not self.case_sensitive:
             self.inventory = [c.lower() for c in self.inventory]
+        # Remove the indices, they're not part of the text input for the rules
         self.inventory = [re.sub(r"{[0-9]+}", "", x) for x in self.inventory]
+        # Rules with "in": "è|é" have two distinct inputs, split them.
```
joanise, line 72: This approach is still not completely right - for example, `^b` means `b` at the beginning of the string, but this code to assemble the inventory sees the caret as a literal instead. I'm not sure how to fix this problem globally correctly, but this patch specifically address alternations in rules, the way they are used in `mic` and `gla`.

The tokenization inventory is only important for non-letter symbols used as letters - e.g., the apostrophe in `mic`, because the tokenizer assume anything that is defined as a letter in the Unicode standard is part of the word. So, while not perfect nor fully general, this solutions fixes the actual use we have on hand at the moment.

g2p/tests/public/data/__init__.py
```diff
@@ -0,0 +1,63 @@
+"""
```
joanise, line 1: The code here was refactored from other parts of the `tests/` software. While not required for the mic tokenization problem, this refactoring work helped with the testing, so it made sense to include it in this PR.

g2p/tests/public/data/mic.psv
```diff
@@ -1,4 +1,13 @@
 mic|mic-ipa|tiꞌam|tiːɑm
 mic|mic-ipa|Miꞌkmaq|miːɡmɑx
+mic|mic-ipa|mi'kmaq|miːɡmɑx
 mic|mic-ipa|sqolj|əsxoltʃ
-mic|mic-ipa|sq|səx
\ No newline at end of file
+mic|mic-ipa|sq|səx
+
+# Test that ' is tokenized correctly as part of the words.
```
joanise, line 7: Comments and blank lines are now supported in `public/data/*.?sv` test data files. I think it'll be nice to document why some weird test cases are being added.

g2p/mappings/tokenizer.py
roedoejet: If ^ is a legitimate character in the orthography, then `escape_special` should be set to `true`, otherwise we could safely strip special regex characters and not treat them as part of the inventory right? the only case where we wouldn't strip is in alternations which you've handled here. So for ^ or $ or other things, if `escape_special` is false, we could just strip them out. What do you think of that @joanise ?

roedoejet (changes requested):
I approve this PR, but it might be nice to address tokenization issues with other special characters (other than `|`)

Thanks again @joanise !

joanise:
@roedoejet That sounds like better logic, I agree. I'm still not fully satisfied with my approach, because we just know there will be more corner cases. This patch really only fixes the cases occurring in current mappings.

I think a much better solution would be to rewrite my tokenizer from scratch using a letter inventory for the language instead, but for now let's just merge this in.

I'll turn your suggestion for caret into an issue so we remember the idea, but it won't be a priority just now.

g2p/mappings/tokenizer.py
joanise: That sounds like better logic, I agree. I'm still not fully satisfied with my approach, because we just know there will be more corner cases. This patch really only fixes the cases occurring in current mappings.

I think a much better solution would be to rewrite my tokenizer from scratch using a letter inventory for the language instead, but for now let's just merge this in.

I'll turn your suggestion for caret into an issue so we remember the idea, but it won't be a priority just now.

PR #191, opened 2022-09-12
ci: adjust coverage configuration for CI and manual testing

joanise:
This PR changes the code coverage configuration to ignore the `tests/` directory, since what we want to test and measure coverage on is the `g2p` module itself, not its testing framework. This is also what we already do in ReadAlongs/Studio, but here 

Also use `.coveragerc` files to get the same config in CI and when interactively running the coverage report. (Although codecov.io and `coverate report` still don't calculate the percentage in exactly the same way, at least now they're doing it on the same hit/miss/partial counts.)


roedoejet (approved):
Makes sense - thanks @joanise !

PR #193, opened 2022-10-07
ci: skip CI tests for commits with #no-ci in the message

joanise:
This is just a minor convenience thing I added to Studio CI workflow as well.

When you're doing work in progress and just want to push a backup on a branch you're planning to throw away, add #no-ci to the commit log and CI gets skipped.


roedoejet (approved):
Oh - that's super cool! Thanks Eric. I'll use that.

Issue #188, opened 2022-08-26
moh -> moh-ipa does not work in studio


roedoejet:
this doesn't work in studio. I think we ought to re-write the test suite so that every lang is also checked through the studio mock as well

joanise:
What happens when you try to use it in Studio?

roedoejet:
No output appears. Looks like a malformed rule error is thrown

PR #194, opened 2022-10-12
Dev.studio test

roedoejet:
Hey Eric,

This is the PR that will fix https://github.com/roedoejet/g2p/issues/188 among a variety of other undiagnosed issues with haa, mic, tli, ctp, and crk

Basically:

- E2E tests are added for studio that run separately (as they require 3.8)
- G2P Studio was out of date with respect to the `rule_ordering` api
- The "reverse" checkbox should not be checked by default in studio, even if it is in the mapping declaration, because mappings are already reversed when they are sent to the Studio
- We actually had a more general bug in g2p - abbreviations which reference other abbreviations (for example defining `VOWELS` as being comprised of `HIGH_VOWELS` and `LOW_VOWELS` abbreviations was broken. This fixes that by recursively expanding abbreviations (with a fail-safe for self-referential infinite recursion errors)
- `prevent_feeding` on individual rules was not working properly because that information was lost when sent to studio. that is now fixed.

Since Studio builds from `master` I'm thinking either we should change it to build from `release` or we should run this tests on every push to `master`.

Thanks!

P.S After this, let's definitely make a release

### DIFF
--- a/.github/workflows/studio-release-tests.yml
+++ b/.github/workflows/studio-release-tests.yml
@@ -1,9 +1,11 @@
 name: Run Tests for Studio
 on:
   pull_request:
-    branches: [release]
+    branches: [release, master]
+  push:
+    branches: [release, master]
 jobs:
-  test:
+  test-studio:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -15,10 +17,10 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip3 install -r requirements.txt
-          pip3 install -r requirements.test.txt
+          pip3 install -r requirements/requirements.test.txt
           pip3 install -e .
-      - name: Ensure browsers are installed
-        run: python -m playwright install --with-deps
+      - name: Ensure browser is installed
+        run: python -m playwright install --with-deps chromium
       - name: Run tests
         run: |
           gunicorn --worker-class eventlet  -w 1 g2p.app:APP --no-sendfile --bind 0.0.0.0:5000 --daemon
--- a/.github/workflows/studio-release-tests.yml
+++ b/.github/workflows/studio-release-tests.yml
@@ -1,0 +1,32 @@
+name: Run Tests for Studio
+on:
+  pull_request:
+    branches: [release]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.6"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip3 install -r requirements.txt
+          pip3 install -r requirements.test.txt
+          pip3 install -e .
+          pip3 install coverage coveralls
+      - name: Ensure browsers are installed
+        run: python -m playwright install --with-deps
+      - name: Run tests
+        run: |
+          gunicorn --worker-class eventlet  -w 1 g2p.app:APP --no-sendfile --bind 0.0.0.0:5000 --daemon
+          sleep 5
+          coverage run run_tests.py release-only
+          coverage xml
+          if git status | grep -E 'static.*json|mapping.*pkl'; then echo 'g2p databases out of date, please run "g2p update" and commit the results.'; false; else echo OK; fi
+      - uses: codecov/codecov-action@v2
+        with:
+          fail_ci_if_error: true # optional (default = false)
--- a/.github/workflows/studio-release-tests.yml
+++ b/.github/workflows/studio-release-tests.yml
@@ -10,23 +10,17 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.6"
+          python-version: "3.8"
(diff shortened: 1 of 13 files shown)


joanise:
@roedoejet I'd like to run the `test_studio.py` test suite locally on my machine, but `gunicorn` is a *nix only piece of software, it doesn't exist for Windows. I'm trying to use `uvicorn` instead but without luck so far: `PRODUCTION= uvicorn  --port 5000 --reload g2p.app:APP` does *something*, but trying to use it just gives me errors http://localhost:5000 and http://localhost:5000/api/v1 both give me "Internal server error" with this stack trace in the console:
```
ERROR:    Exception in ASGI application
  File "C:\Users\joanise\Miniconda3\envs\ras-py38\lib\site-packages\uvicorn\protocols\http\h11_impl.py", line 403, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "C:\Users\joanise\Miniconda3\envs\ras-py38\lib\site-packages\uvicorn\middleware\proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "C:\Users\joanise\Miniconda3\envs\ras-py38\lib\site-packages\uvicorn\middleware\asgi2.py", line 19, in __call__
    instance = self.app(scope)
TypeError: __call__() missing 1 required positional argument: 'start_response'
ERROR - Exception in ASGI application

  File "C:\Users\joanise\Miniconda3\envs\ras-py38\lib\site-packages\uvicorn\protocols\http\h11_impl.py", line 403, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "C:\Users\joanise\Miniconda3\envs\ras-py38\lib\site-packages\uvicorn\middleware\proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "C:\Users\joanise\Miniconda3\envs\ras-py38\lib\site-packages\uvicorn\middleware\asgi2.py", line 19, in __call__
    instance = self.app(scope)
TypeError: __call__() missing 1 required positional argument: 'start_response'
INFO:     127.0.0.1:59759 - "GET /api/v1 HTTP/1.1" 500 Internal Server Error
```
The stack trace is the same regardless of the URL I try to fetch.

Any change you're able to launch it with uvicorn on your Mac and could share the command that works?

.github/workflows/studio-release-tests.yml
```diff
@@ -0,0 +1,26 @@
...
+      - uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.8"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip3 install -r requirements.txt
+          pip3 install -r requirements.test.txt
```
joanise, line 18:
```suggestion
          pip3 install -r requirements/requirements.test.txt
```

joanise:
I'd also like to create a branch where we relax the studio test CI workflow to run on any push, so we can see it in action and hopefully working.

g2p/static/custom.js
```diff
@@ -223,7 +226,7 @@ function createTable(index, data) {
     let headerLabels = headers.map(header => header.replace('_', ' ').split(' ').map(x => { return x.charAt(0).toUpperCase() + x.slice(1) }).join(' '))
     var hotSettings = {
         data: data,
-        columns: headers.map(x => { return { 'data': x, type: 'text' } }),
+        columns: headers.map(x => { if (x !== "prevent_feeding") { return { 'data': x, type: 'text' } } else { return { 'data': x, type: 'checkbox' } } }),
```
joanise, line 229: I'm used to our black formatting in Python converting such lines to multi-line. Are there similar conventions for JS or is that more free-for-all?

JS has so many nastily nested braces and parens, I would not want something anywhere as string as black for it, but I guess intuitively I have have formatted this code like this, although I'm not really satisfied with this option either: 

```
columns: headers.map(x => { if (x !== "prevent_feeding") { 
             return { 'data': x, type: 'text' }
         } else {
             return { 'data': x, type: 'checkbox' }
         }}),
```

Don't take this as a real suggestion, I don't actually like it! I'm just discussing this as part of my learning JS.

joanise:
> I'd also like to create a branch where we relax the studio test CI workflow to run on any push, so we can see it in action and hopefully working.

Done: see branch dev.studio-ci-test and https://github.com/roedoejet/g2p/actions/runs/3236397077 for one run.
There is exactly one mapping error, and a strange one at that:
```
WARNING - test_langs.py: mapping error: eyewayawoyoeëäwä from haa to haa-ipa should be eieoæiæooioeəaoɑ, got EH IH EH AO AE IY AE AO AO IY AO EH AH AE AO AA
```
why did we get ARPABET output when asking for `haa-ipa`?

joanise:
About running with uvicorn, I feel silly... A simple `python run_studio.py` at the root did the trick and I'm running the test suite on my machine now. Better speed than what you first told me about, with 7 minutes to run the whole suite. Still slow, but manageable!

joanise (approved):
Great work. It's good to have the g2p CLI and the g2p-studio app align again, and it'll be great to have that `test_studio.py` test suite make sure we catch future changes that break that.

PR #198, opened 2022-10-14
test: add --describe option to run.py and exit 1 on error

joanise:
I found it hard to manually inspect `run.py` to determine what tests `python run.py dev` might have left out. So I automated it.

`python run.py --describe dev` now lists what is include and, most importantly, what test cases are excluded.

Also, make the different ways to run the test suite consistently exit 1 on error.


joanise:
This PR is messy. I'll reorg a bit and resubmit later.

PR #201, opened 2022-11-01
change(mappings): add additional fn_unicode PUA mappings

eddieantonio:
While looking at https://github.com/portagenetwork/roadmap/issues/239, we found there are more characters in FNuni_v2.3 that can be mapped from their current assignments in the BMP private use area to conventional Unicode character sequences.

~~I'm a little bit iffy about them right now, but I'm starting the pull request now if you have any questions.~~


g2p/mappings/langs/font-encodings/fn_unicode.csv
```diff
@@ -1,21 +1,40 @@
+\uF000,x\u030c
 \uF101,c\u0313
-\uF102,č\u0313
+\uF102,c\u030c\u0313
```
eddieantonio, line 3: I used fully NFD forms here instead of the half-composed, half-decomposed value that was here before.

g2p/mappings/langs/font-encodings/fn_unicode.csv
```diff
@@ -1,21 +1,40 @@
...
 \uF103,g\u0313
+\uF104,j\u030c\u0313
 \uF105,k\u0313
 \uF106,l\u0315
 \uF107,m\u0313
 \uF108,n\u0313
 \uF109,ŋ\u0313
-\uF110,z\u0313
-\uF111,\u019b\u0313
```
eddieantonio, line 10: These guys were in ASCIIbetical order, whereas the table I've contributed to is ordered by code point.

g2p/mappings/langs/font-encodings/fn_unicode.csv
```diff
@@ -1,21 +1,40 @@
...
 \uF10B,q\u0313
 \uF10C,r\u0313
 \uF10D,t\u0315
 \uF10E,w\u0313
 \uF10F,y\u0313
-\uF132,\u1DBF
+\uF110,z\u0313
+\uF111,ƛ\u0313
+\uF112,ɣ\u0313
+\uF113,ʕ\u0315
```
eddieantonio, line 20: I'm unsure about this one. Here is the glyph from the font:

Screengrab of pharyngeal_glottalized from FNuni_2.3

All other characters in this series combine a base letter with [◌̓ U+0313  COMBINING COMMA ABOVE](https://www.compart.com/en/unicode/U+0313), however I have opted for [◌̕ U+0315  COMBINING COMMA ABOVE RIGHT](https://www.compart.com/en/unicode/U+0313) for this one.

Compare:

| Combing character | Result |
|-|-|
| U+0313 | ʕ̓ |
| U+0315 | ʕ̕ |

eddieantonio (commented):
Please let me know if things look right here

g2p/mappings/langs/font-encodings/fn_unicode.csv
joanise: Makes it easier to read, good idea.

g2p/mappings/langs/font-encodings/fn_unicode.csv
joanise: This should make no functional difference in the end, since the mapping uses NFD normalization (which is applied to rules as well as inputs), but if you find it easier to read this way it's fine.

g2p/mappings/langs/font-encodings/fn_unicode.csv
joanise: Based on inspecting the image at https://fnel.arts.ubc.ca/resources/font/ I have to agree with your choice. This is the character you get with Shift+Option 9, right? It's an odd inconsistency, but if it's what's been agreed on we should not deviate.

joanise (commented):
Some comments interspersed, but fundamentally you're the expert on this, and I assume you're making sure the mapping is precise.

I was going to suggest adding unit testing for it in `/tests/public/data`, but it's really just a straightforward mapping from single characters out, so there's not much value in doing so. There are not "tricky" cases to exercise.

Question: is this g2p mapping being used in a project you're working on at the moment? I'd love know more about it.

joanise:
Most importantly, thanks for this contribution!

eddieantonio:
I have manually reviewed every character on the FNUni_v2.3 keyboard for macOS and verified that the mapping does the right thing for each possible keypress. The only character that is weird is U+F204, which in the font is a superscript square root symbol (like √, but as a superscript). **I cannot find a suitable Unicode mapping for this character**, but it seems that this mapping already mapped it to a regular √. Besides, it can be emulated by typesetting that character using a word processor's superscript, e.g., \<<sup>√</sup>\>.

Other than that, I think this mapping is good to go!

@roedoejet, I'd appreciate it your comments on this!

eddieantonio:
@joanise 
> Question: is this g2p mapping being used in a project you're working on at the moment? I'd love know more about it.

Not a project I'm working on, but to address https://github.com/portagenetwork/roadmap/issues/239. Someone who works on it contacted me out of the blue, and I knew it was a problem that g2p could solve!

g2p/mappings/langs/font-encodings/fn_unicode.csv
roedoejet: Agreed. I have a mild preference for the half composed/half decomposed actually since the mapping takes care of normalization and it's a bit more readable, but I don't really care

g2p/mappings/langs/font-encodings/fn_unicode.csv
roedoejet: yep good idea

g2p/mappings/langs/font-encodings/fn_unicode.csv
roedoejet: Yea I feel like for line height reasons some of these tall characters opted for putting the comma above right. This sounds reasonable. `t` also does it strangely though

roedoejet:
> I have manually reviewed every character on the FNUni_v2.3 keyboard for macOS and verified that the mapping does the right thing for each possible keypress. The only character that is weird is U+F204, which in the font is a superscript square root symbol (like √, but as a superscript). **I cannot find a suitable Unicode mapping for this character**, but it seems that this mapping already mapped it to a regular √. Besides, it can be emulated by typesetting that character using a word processor's superscript, e.g., <√>.
> 
> Other than that, I think this mapping is good to go!
> 
> @roedoejet, I'd appreciate it your comments on this!

Yes - I believe this is just used for when linguists want to annotate that a particular morpheme is a root - it's honestly probably not used very often, and isn't part of an actual community orthography. The superscript was probably to make it seem more obviously like an annotation instead of a segment. Anyways, as you point out if someone really wants to make it superscript they can in their word processor!

roedoejet (approved):
Looks great - thanks for filling this out @eddieantonio ! It wasn't finished before and now it's nice and complete 😊

roedoejet:
Just FYI @eddieantonio - let us know if your group needs a release soon. They can of course build from the `master` branch for the recent updates, but we just had a release a couple weeks ago so aren't planning on releasing to pypi for another couple months. If this is inconvenient for your group let us know and we can release a patch sooner. However, the studio: https://g2p-studio.herokuapp.com/ updates on every push to `master` so that is up and running and they can also use the API, see the interactive docs here: https://g2p-studio.herokuapp.com/docs and an example here: https://g2p-studio.herokuapp.com/api/v1/g2p?in-lang=fn-unicode-font&out-lang=fn-unicode&text=%5CuF113

eddieantonio:
@roedoejet No need for a release immediately! That said, I _might_ have found an issue in this mapping I didn't catch earlier. I'll get back to you soon.

joanise:
> Not a project I'm working on, but to address [portagenetwork/roadmap#239](https://github.com/portagenetwork/roadmap/issues/239). Someone who works on it contacted me out of the blue, and I knew it was a problem that g2p could solve!

Ah, nice. It's good to see this tech branch out like this, and I'm glad you were there to recognize it could fill the gap.

PR #202, opened 2022-11-04
fix(mappings): fn_unicode bullet operator → middle dot

eddieantonio:
This is a follow up to #201, but it's a bit more controversial.

On the [Mac version](http://fnel2020.sites.olt.ubc.ca/files/2015/09/FN-UNI2-Mac-keyboard.gif) of [FNUni](https://fnel.arts.ubc.ca/resources/font/), Shift+Option+3 inserts ∙ U+2219 BULLET OPERATOR. However, I think the intended character is · U+00B7 MIDDLE DOT. You can hopefully see why I overlooked this detail when I was preparing #201!

Evidence for prefering U+00B7 · U+2219 MIDDLE DOT:

 - U+00B7 is used the [vowel length mark in Makah](https://makahmuseum.com/departments/makah-language-program/)
 - When working on the Makah keyboard, I had an exchange with Craig Cummings who explicitly recommended U+00B7 with the following supporting details:
     - p.6 "But What's the Harm, Anyway?" of https://www.unicode.org/L2/L2012/12361-n4340.pdf
     - U+00B7 collates as a "word" character https://www.unicode.org/wg2/docs/n4339.pdf
     - (both documents seem to promote  · U+00B7 MIDDLE DOT for Americanist transcription systems)
 - The existing mapping already has a rule for ∛ to U+00B7
 - Word segmentation algorithms will break words at U+2219, but consider U+00B7 to be part of the word

I assume some sort of middle dot looking character is included in the font/keyboard as a morpheme separator (@roedoejet, does this sound reasonable?), but it's also used in Makah as a "letter". Distressingly, Option+3 inserts • U+2022 BULLET, so that means the keyboard inserts two middle dot looking things.

---

I'm having difficulty coming up with any genuine test cases -- do you know where I can find genuine text that would be plausibly be written with FNUni?


joanise:
Confusing: U+2219 BULLET OPERATOR vs U+2219 MIDDLE DOT have the same code point but a different description. I think you're being precisely correct but can you confirm? If I understand correctly the question what should U+2219 get mapped to in the end.

eddieantonio:
> Confusing: U+2219 BULLET OPERATOR vs U+2219 MIDDLE DOT have the same code point but a different description. I think you're being precisely correct but can you confirm? If I understand correctly the question is: what should U+2219 get mapped to in the end?

😅 that was me being confused with copy-paste! MIDDLE DOT, the proposed replacement, should be U+00B7!

So the question is whether remapping ∙ U+2219 BULLET OPERATOR to  · U+00B7 MIDDLE DOT. The keyboard inserts  ∙ U+2219, but I believe  · U+00B7 is semantically intended.

joanise:
Ah, OK, that makes more sense now. Thanks for the clarification.

I'll let @roedoejet review the substance here, because I have no expertise to judge whether this mapping is appropriate or not. But thanks for adding a test file in public/data, I'm always happier have those there.

joanise:
Coverage analysis for this repo only analyses code coverage. It would be nice here to have a second report that counts the proportion of g2p rules exercised by the test suites are shows which ones are not. But boy, that would be a big project.

roedoejet:
> I'm having difficulty coming up with any genuine test cases -- do you know where I can find genuine text that would be plausibly be written with FNUni?

Yes well the only place I really know is for hən̓q̓əmin̓əm̓ but I don't believe they have any publicly available content, so we might have to just have our silly tests in the meantime. From my perspective the reasons you give for supporting U+00B7 are very clear and I accept this change, even though it means we might be unilaterally normalizing away some distinction. I think for the purposes of this library this is good; the word segmentation argument alone is pretty compelling. Thanks @eddieantonio !

roedoejet (approved):

PR #206, opened 2022-11-21
Fix all CI deprecation warnings

joanise:
Hard to test until we do the next release, but carefully reviewed.

I suggest we merge this in now, but monitor closely when we make our next release March 1.

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -47,10 +47,8 @@ jobs:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         custom_tag: ${{ steps.determine_tag.outputs.TAG_VERSION }}
     - name: Create a GitHub release
-      uses: actions/create-release@v1
-      env:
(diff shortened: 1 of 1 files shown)


roedoejet:
see my note here: https://github.com/ReadAlongs/Studio/pull/134#issuecomment-1340208665 

Especially considering these changes affect two release actions, I think we need to test this on a separate repo.

joanise:
Yeah, agreed here too.

joanise:
@roedoejet I have now tested this in a separate repo, and I can confirm that it is safe to merge.

While I was there, I have tested what happens if you make a release and delete it, should you be unhappy with it. The results are nice.
 - I deleted the release and the tag associated with it
 - I pushed a new release to my test repo
 - The new release had all the update messages from the commits that were included in the deleted commit, going right back to the previous commit that was still there.

I recommend merge this into `main` now.

roedoejet (approved):
Great, thanks for testing @joanise !

PR #209, opened 2022-12-22
Misc improvements to the g2p-studio static app

joanise:
- add padding at the very bottom, otherwise it looks like you can't scroll to the end of the abbreviations table
 - have `convert()` send only non-empty rules and abbreviations to the server, since they just case lots of spurious warnings, and the empty lines in g2p-studio are there to facilitate insert rules, they're not meant to actually be empty rules
 - label the first column of the abbreviations table "Abbreviation Name"
 - remove the dead `"export-abbs"` listener that depended on the `varhot` variable which got removed a while ago (we now use `$('#export-abbs').click(...)` instead).

Still TODO because I could not figure it out: when you click on EXPORT for either Custom Rules or Custom Abbreviations, rules with an empty `.in` and abbreviations with an empty name should be filtered out too, before the `.csv` files are generated and downloaded.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,7 +4,7 @@ on:
   - pull_request
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     # #no-ci in the commit log flags commit we don't want CI-validated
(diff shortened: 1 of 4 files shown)


roedoejet (approved):
Great changes Eric - this was unnecessary noise.

PR #207, opened 2022-11-25
ci: add CodeQL automated vulnerability scanning

joanise:

PR #210, opened 2023-01-06
removed an unnecessary mapping

deltork:
removed a mapping that was causing issues


joanise:
@deltork  You're the expert on this, so I'll approve. But please run `g2p update` once more, CI is complaining about that.

joanise:
@deltork Nevermind my comment, there's something wrong with the autobuild. Even a patch that does not touch the mappings at all is getting is the same warning about needing to rerun `g2p update`. I think some dependencies have jumped too far forward and we need to lock them back, but I don't know which one(s) yet.

@roedoejet I'm analyzing this problem, which is also affecting #207 CI and will probably block and any and all CI until we figure it out.

joanise (approved):

PR #212, opened 2023-01-10
feat: language name for spelling variants describe the variant

joanise:
This change is potentially more controversial, so I want to submit it for more careful review.

Basically, I changed "Inuktitut" to "Inuktitut Syllabics" and "Inuktitut Romanized" and same for kwk and oji mappings where there are orthographic variants we're supporting.

The idea is so that Studio-Web would display something more informative in the list of languages. There might be other solutions, though, so please comment!


dhdaines (approved):
This doesn't seem very controversial to me.  We could be more explicit that "language" really means "writing system" (most people do not really make this distinction, but that's not our fault) but this clarification is necessary where there are mutliple current orthographies.

roedoejet (approved):
Seems sensible to me.

PR #218, opened 2023-02-21
Fix for Arabic text

littell:
There were two errors caused by Arabic text, leading to 422 and 500 errors in Studio Web when certain Arabic words were pasted in.

(1) Letters in presentation forms (e.g. U+FEB7 ARABIC LETTER SHEEN INITIAL FORM) isn't understood by text-unidecode.  Getting rid of the positional information requires a K normalization, like NFKC.

(2) If text-unidecode turns a character to a non-alpha character, this causes trouble downstream.  Many Arabic letters result in non-alpha characters.  We used to strip these out, but this got lost in the port from Studio to G2P.  I've added it back in.

- [ ] Mapping files are added in `g2p/mappings/langs`
- [ ] Mapping is either added to an existing folder or a new folder has been added
- [ ] Language folder and files use appropriate [ISO 639-3 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-3_codes)
- [ ] Config.yaml file includes all author names, and settings necessary
- [ ] Please add some test data in `g2p/tests/public/data`. The added file should be a csv/tsv/psv file and each row should have the format `[input_mapping_code,output_mapping_code,input_string,output_string]`
- [ ] As the last step, G2P has been updated by running `g2p update` locally and committing the change
- [ ] You agree to license your contribution under the same license as this project (see [LICENSE](https://github.com/roedoejet/g2p/blob/master/LICENSE) file).

* **Other information**:

### DIFF
--- a/g2p/tests/test_unidecode_transducer.py
+++ b/g2p/tests/test_unidecode_transducer.py
@@ -41,6 +41,16 @@ class UnidecodeTransducerTest(TestCase):
             tg.output_string, "EY T EY  N UW N AA V UW T  N OW N AA F OW T "
         )
 
+    def test_unidecode_arabic_to_arpabet(self):
+        transducer = make_g2p("und", "eng-arpabet")
+        tg = transducer("السلام عليكم")
+        self.assertEqual(tg.output_string, "L S L M L Y K M ")
+
+    def test_unidecode_arabic_presentation_to_arpabet(self):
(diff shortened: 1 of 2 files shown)


littell:
Oh, just realized I messed up some tests by inadvertantly removing spaces.  I'm closing this and reopening it with the fix.

PR #219, opened 2023-02-21
Fix for processing Perso-Arabic scripts with "und"

littell:
Some Perso-Arabic inputs were leading to 422 and 500 errors in Studio Web, due to:

(1) Arabic presentation forms (e.g., those that have positional information encoded) aren't handled by text-unidecode; we have to do a K normalization (e.g. NFKC) before sending them to unidecode.

(2) Some words were unidecoding into non-alpha characters that we weren't expecting, like backtick.  We could have just handled this one character... but in the bigger picture, we don't 100% know what characters can come out of unidecode and what they "mean", and even one unexpected character in the result can cause a 422.  So to prevent this kind of issue in the future, I added a filter that replaces unexpected characters with ''.

### DIFF
--- a/g2p/transducer/__init__.py
+++ b/g2p/transducer/__init__.py
@@ -525,8 +525,10 @@ class Transducer:
         tg = TransductionGraph(to_convert)
 
         # Conversion is done character by character using unidecode
-        converted = [unicodedata.normalize("NFKC", c) for c in to_convert]
-        converted = [text_unidecode.unidecode(c) for c in converted]
+        converted = [
+            text_unidecode.unidecode(unicodedata.normalize("NFKC", c))
+            for c in to_convert
+        ]
         converted = [
             c if c.isalpha() or c in UNIDECODE_SPECIALS else "" for c in converted
         ]


joanise (approved):
Looks clean, and works well. Thanks!

Thanks for unit testing.

g2p/transducer/__init__.py
```diff
@@ -522,7 +525,11 @@ def apply_unidecode(self, to_convert: str):
         tg = TransductionGraph(to_convert)
 
         # Conversion is done character by character using unidecode
-        converted = [text_unidecode.unidecode(c) for c in to_convert]
+        converted = [unicodedata.normalize("NFKC", c) for c in to_convert]
```
roedoejet, line 528: maybe change to `[text_unidecode.unidecode(unicodedata.normalize("NFKC", c)) for c in to_convert]` just to save the extra loop

roedoejet (approved):
Looks great! one minor perf suggestion but this is awesome, thanks for fixing so quickly @littell - I'm kind of surprised readalongs will just work seemingly with rtl and arabic after this.

g2p/transducer/__init__.py
joanise: I don't think it's equivalent. I had the same thought, but NFKC can expand a char in several chars, then you loop again after expansion.

joanise: You could do it in nested loops, but this is clearer I think.

joanise:
@roedoejet Commits are not conventional, can we make the merge commit say `fix: blah blah` instead of Merge, to get the desired change log?

joanise:
Simpler, I'll just rewrite the last commit message and force push it.

g2p/transducer/__init__.py
roedoejet: `converted` will still be a list of strings though, regardless of whether `unicodedata.normalize("NFKC", c)` returns a different length string. So, I think they should actually be equivalent.

joanise: Sorry, that was nonsense, since we're working with arrays, we're not joining in between.

So yes, the three loops could be converted to a single, equivalent, loop, it's true.

roedoejet: no, just the first 2, since the 3rd has a conditional dependent on the output of the first two transformations.

PR #225, opened 2023-02-27
Fix to Han/Kanji characters in Und

littell:
Fixes to how we sanitize unidecode outputs so that we don't inadvertently throw out the outputs of Han/Kanji unidecoding.  Same changes as the last commit, but on a clean branch.


joanise (approved):
This makes sense to me. I vaguely wondered what would happen a character "multiplied", but I had naively assumed your Arabic test exercised that. Obviously not!

PR #228, opened 2023-03-04
Reimplement G2P API and Studio with FastAPI

dhdaines:
FastAPI has no built-in server, so `g2p run` can no longer exist (unless we make it just call uvicorn, which I suppose we can do)

On the bright side, no more Werkzeug... but I haven't tested this with Python 3.6 yet.


roedoejet:
I'll have a look at this next week! but since everything is already implemented in flask & swagger & socketio, what's the reason for changing to fastapi (other than that fastapi is a lot of fun)? I imagine re-writing all these tests will be a bit annoying

dhdaines:
> I'll have a look at this next week! but since everything is already implemented in flask & swagger & socketio, what's the reason for changing to fastapi (other than that fastapi is a lot of fun)? I imagine re-writing all these tests will be a bit annoying

Hmm, good point!  I think it was more interesting as a learning exercise than anything.  It makes the API code a bit simpler, which is why I did it, since I wanted to add a new API, but it won't be a problem to go back and do it with Flask.

Unfortunately it doesn't seem to work to do the API in FastAPI and the rest with Flask due to all the fancy stuff we do with Flask - the WSGI middleware can't really handle it.

dhdaines:
I'm going to keep this around as an experimental branch but don't expect it to be merged!

roedoejet:
I'm going to close this since it's an experimental branch.

PR #229, opened 2023-03-04
fix: update requirements.txt for python 3.6

dhdaines:
The Numpy version is incompatible with panphon but seems to install anyway (more recent versions demanded by panphon are not available for 3.6).  editdistance 0.6.2 won't compile properly so back it off to 0.6.1

I was worried about playwright not working on 3.6 but I see that we use 3.8 to run that test in any case.

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -10,7 +10,9 @@ flask-talisman>=0.7.0
 flask-restful>=0.3.9
 networkx==2.5
 panphon>=0.19
-python-engineio==3.14.2 
+editdistance==0.6.1
+numpy==1.19.5
+python-engineio==3.14.2
 python-socketio==4.6.1
 pyyaml>=5.2
 regex


joanise:
I don't quite understand why we want to do this. Without it, pip was installing more recent versions of some of these things when possible, and they were working fine, on my Windows machine anyway. Unless there is a compelling reason to prevent that, I would let pip keep the more recent version when it can.

In Python 3.8, this resulted in my Panphon installation being downgraded to 0.19.0 instead of the 0.20.0 I had, and editdistance from 0.6.2 down to 0.6.1, whereas in Python 3.6 I already had the lower versions. And things worked in both cases.

What's the context where not pinning these actually caused a problem?

dhdaines:
It doesn't install on Python 3.6 on my machine (with miniconda), but if it works in the CI then I guess we should just leave it as is.  Closing this.

PR #231, opened 2023-03-05
New API suitable for subword alignment applications

dhdaines:
This API adds the missing functionality needed to do subword-level alignment. Of note:

- It returns only minimal monotonic substring alignments by default, as this is the most useful thing for apps that want to use subword alignments to refer back to the input text.
- You can request that the transductions returned be composed starting at a given step, which is what you want if you wish to recover the original language IPA from ARPABET output, for instance.
- Tokenization is done by default (but can be turned off), and separate conversion lists are returned for each token.  Non-converted text (e.g. punctuation) is returned as a single conversion with `in_lang` and `out_lang` as `None`, so you can tell it isn't converted
- If numerical indices are requested, then the input and output nodes are returned explicitly as lists to avoid issues with {Java, JavaScript, C#, etc, etc} which have a different notion of "character" than {Python, Rust, etc, etc} making these indices possibly invalid when used to index into strings.
- `ancestors` and `descendants` have more meaningful names now I think (`inputs_for` and `outputs_for`)
- there is a `path` endpoint to just find the path (if it exists)
- since we do tokenization, etc, the input could potentially be quite long, so we use POST not GET
- the `/langs` endpoint is much more user-friendly, returning languages and names by default, or all nodes if you really want it to

### DIFF
--- a/g2p/api_v2.py
+++ b/g2p/api_v2.py
@@ -37,9 +37,9 @@ from networkx.algorithms.dag import ancestors, descendants
 from pydantic import BaseModel, Field
 
 import g2p
+import g2p.mappings.langs as g2p_langs
 from g2p.exceptions import InvalidLanguageCode, NoPath
 from g2p.log import LOGGER
-from g2p.mappings.langs import LANGS_NETWORK
 from g2p.transducer import (
     CompositeTransductionGraph,
     TransductionGraph,
@@ -81,19 +81,50 @@ app.add_middleware(
     CORSMiddleware, allow_methods=["GET", "POST", "OPTIONS"], **middleware_args
 )
 
-
-# Get the langs
-LANGS = sorted(LANGS_NETWORK.nodes)
-LanguageNode = Enum("LanguageNode", [(name, name) for name in LANGS])  # type: ignore
+# All possible language codes
+LanguageNode = Enum("LanguageNode", [(name, name) for name in g2p_langs.LANGS_NETWORK.nodes])  # type: ignore
 
 
-@api.get("/langs", response_description="List of supported language code strings")
-def langs() -> List[str]:
-    """Return list of supported language codes.  Note that these are not
-    exactly *languages* but rather writing or phonetic systems
-    associated with a given language.
+class SupportedLanguage(BaseModel):
+    """Writing or phonetic system for conversion"""
+
+    code: str = Field(
+        description="Language, writing, or phonetic system code as passed to /convert",
+        example="eng",
+    )
+    name: Union[str, None] = Field(
+        description="Display name (not internationalized) for language, may be None",
+        example="English",
+    )
+
+
+@api.get(
+    "/langs",
+    response_description="Supported writing or phonetic systems for conversion",
+)
+def langs(
+    allnodes: bool = Query(
+        False,
+        description="Return all nodes in the conversion network rather than just top-level languages.",
+    )
+) -> List[SupportedLanguage]:
+    """Return a list of language codes and their names.  If `allnodes` is
+    given, return all nodes in the conversion network (which may or
+    may not correspond to languages).
     """
-    return LANGS
+    if allnodes:
+        return [
+            SupportedLanguage(
+                code=code, name=g2p_langs.LANGS.get(code, {}).get("language_name", None)
+            )
+            for code in sorted(g2p_langs.LANGS_NETWORK.nodes)
+        ]
+    else:
+        return [
+            SupportedLanguage(code=code, name=g2p_langs.LANGS[code]["language_name"])
+            for code in sorted(g2p_langs.LANGS.keys())
+            if "language_name" in g2p_langs.LANGS[code] and code != "generated"
+        ]
 
 
 class ConvertRequest(BaseModel):
@@ -295,7 +326,7 @@ def outputs_for(
     are all the phonetic or orthographic systems into which you can convert
     this input.
     """
-    return sorted(descendants(LANGS_NETWORK, lang.name))
+    return sorted(descendants(g2p_langs.LANGS_NETWORK, lang.name))
 
 
 @api.get(
@@ -309,7 +340,7 @@ def inputs_for(
     are all the phonetic or orthographic systems that you can convert
     into this output.
     """
-    return sorted(ancestors(LANGS_NETWORK, lang.name))
+    return sorted(ancestors(g2p_langs.LANGS_NETWORK, lang.name))
 
 
 @api.get(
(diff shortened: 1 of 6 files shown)


joanise:
It's gonna take me a while to review this PR, but a question just based on the PR description. You say the substring alignments are guaranteed to be monotonic, but our rules allow indices to specify non-monotonicity explicitly. How are you handling that now?

dhdaines:
Note that also I didn't notice that we do allow insertions and deletions in the indices, so I need to handle that better.

dhdaines:
> OK, I'll review it later. But yeah, swaps, insertions, deletions, those are all possible parts of the system. String indices have their problems in some cases, but are super flexible. That's probably a double-edged sword, though...

Yes - I wonder if, as @littell mentions, we should define indices to be in terms of Unicode grapheme clusters, as Swift does:

https://pypi.org/project/grapheme/
https://docs.swift.org/swift-book/documentation/the-swift-programming-language/stringsandcharacters/#Extended-Grapheme-Clusters

joanise:
Right, using graphemes would make good logical sense. We get partly at that when we make a g2p mapping use NFC, where graphemes and characters are closer to a 1-1 correspondance (but not always, since not all possible graphemes have a single character version). Doing indices over graphemes would normalize indices in a nice way, making them semantically the same regardless of NFC or NFD normalization settings.

I was not familiar with that `grapheme` project, it sounds very useful. I really like the concept.

The changes might go a little deep, though, because anything using these indices would need to be patched to interpret them as indices over graphemes. We'd have to be thorough looking for any and all code using the indices. So that would definitely be a breaking change in the g2p library.

Also, currently we allow things like saying use NFD and then map ACUTE ACCENT to nothing, so we allow mapping over sub-parts of a grapheme. I don't know if and where we use this, but we support it, so it would be something to look for to make sure we don't break any existing mapping that relies on it. Actually, I know of one case I just wrote recently: `oka_equiv.csv` normalizes diacritics because Nsyilxcen is not always systematic about which exact code point they use for their accents. And I even do exactly the NFD hack I describe above: oka->oka-equiv has norm=NFD, and then oka-equiv->oka-ipa has norm=NFC.

dhdaines:
Note: I merged instead of rebasing from `main` because it was *much* easier... otherwise I had to keep resolve the same conflicts over and over and over.

joanise:
> Note: I merged instead of rebasing from `main` because it was _much_ easier... otherwise I had to keep resolving the same conflicts over and over and over.

Very reasonable.

joanise:
Testing this API this morning. Nice API!

I'd want more documentation of the parameters for `/convert`, though. `Studio/readalongs/web_api.py` is not perfect, but at https://readalong-studio.herokuapp.com/api/v1/docs you can see that we documented all the args and returns etc in the main docstring in `web_api.py` using lists, and the result is fairly legible in the generated documentation.
I would have liked the documentation generated from the Request and Response classes to be it, but I'm not very satisfied with how that's all just dumped at the bottom of the page, and details in there are scant anyway, it's not much more than a list of fields with their types.
Anyway, I feel http://localhost:5000/api/v2/docs could better document the args you can pass in through the POST body and what they do.

You can also set `response_model=<classname>` to document the response schema, like `web_api.py` does for `/assemble` and `/langs`, and I think it would add the schema at the bottom. This is useful to know exactly what fields the results can have, although it still lacks the explanations of what each field means that I would hope for.

This is something about FastAPI I'm less than impressed with: for GET, the parameters get well documented, but for POST not really. So only `/convert` needs extra documentation in this API.

g2p/api_v2.py
```diff
@@ -0,0 +1,406 @@
...
+        allow_origins=[
+            os.getenv("ORIGIN", "https://readalong-studio.mothertongues.org"),
+        ],
+    )
+app.add_middleware(
+    CORSMiddleware, allow_methods=["GET", "POST", "OPTIONS"], **middleware_args
+)
+
+# All possible language codes
+LanguageNode = Enum("LanguageNode", [(name, name) for name in g2p_langs.LANGS_NETWORK.nodes])  # type: ignore
```
joanise, line 82:
```suggestion
LanguageNode = Enum("LanguageNode", sorted((name, name) for name in g2p_langs.LANGS_NETWORK.nodes))  # type: ignore
```
While semantically identical, the `sorted()` means that when you use the "Try it out" button in the API docs, the drop box has languages in order, making them a lot easier to find.
And, the list of available values is also ordered, making it similarly easier to use.

g2p/api_v2.py
```diff
@@ -0,0 +1,406 @@
...
+    are all the phonetic or orthographic systems that you can convert
+    into this output.
+    """
+    return sorted(ancestors(g2p_langs.LANGS_NETWORK, lang.name))
+
+
+@api.get(
+    "/path/{in_lang}/{out_lang}",
+    response_description="Path from {in_lang} to {out_lang}",
+)
```
joanise, line 395: You get slightly better documentation if you add
`response_model=List[str]` to the decorator. I think all the endpoints would get better generated documentation if you specify the `response_model` explicitly. I don't know why the function's `-> List[str]` return type declaraction doesn't already do the same thing!

g2p/api_v2.py
```diff
@@ -0,0 +1,406 @@
...
+    try:
+        transducer = g2p.make_g2p(in_lang, out_lang)
+        if request.tokenize:
+            tokenizer = g2p.make_tokenizer(in_lang)
+            tokens = tokenizer.tokenize_text(request.text)
+        else:
+            tokens = [{"text": request.text, "is_word": True}]
+    except NoPath:
+        raise HTTPException(
+            status_code=400, detail=f"No path from {in_lang} to {out_lang}"
```
joanise, line 296: Pylint suggests doing `except NoPath as exc` and `raise ... from exc`. I still don't fully understand what that changes, but I tend to just do it when it reminds me.

joanise: Ah, wait, reading up on this, the better solution here would be `raise HTTPException() from None` because the caller does not need to know that the cause is our internal `NoPath` exception, they only need to get our 400`.
Informative reading: https://stackoverflow.com/questions/24752395/python-raise-from-usage

joanise:
@dhdaines I just rebased this PR for you, you can hard-reset your branch to `dev.apiv2-rebased`. It was easy once I finally realized there were about a dozen commits that had already been merged into origin/main via #238. The trick was to drop all of those and only keep the ones that are new to this PR.

dhdaines:
> @dhdaines I just rebased this PR for you, you can hard-reset your branch to `dev.apiv2-rebased`. It was easy once I finally realized there were about a dozen commits that had already been merged into origin/main via #238. The trick was to drop all of those and only keep the ones that are new to this PR.

Thanks!

PR #232, opened 2023-03-06
Remove spontaneous extraneous spaces from und-ipa

dhdaines:
This is mainly a problem for Han characters, where unidecode will insert a space between each syllable in the `und-ascii`.  If we only care about the output text for `eng-arpabet` this isn't a huge problem since we will just split on whitespace, *but* if we care about IPA output, or about alignments, it is a problem because it introduces spurious alignments for these phantom spaces, e.g. we get something like this in the `pretty_edges` for `und-ipa -> eng-ipa`:

```
$ g2p convert -e '你好' und eng-arpabet
(... snip ...)
        [   ['n', 'n'],
            ['i', 'i'],
            [' ', ' '],
            ['h', 'h'],
            ['a', 'ɑ'],
            ['o', 'o'],
            ['o', 'ː'],
            [' ', ' ']],
```

Those spaces aren't IPA, and they aren't in the input, we don't want them there.  Instead of expecting the user to special-case them, we should just remove them.


joanise:
This seems _almost_ uncontroversial to me. I'm pretty sure it's the right thing to do, but I wonder about Chinese text, where the are typically no word boundaries. We'll end up with a whole sentence being treated as just one word in Studio.
But I think that's orthogonal to this PR: the spaces unidecode inserts marks character boundaries, but since Studio would have already tokenized before asking g2p to convert, then indeed, those spaces are being inserted at the wrong place, even if maybe they are desirable in other contexts.

dhdaines:
Yeah - ultimately what we want to do is to remove the spaces introduced by Unidecode but leave any others intact. For Chinese and (especially) Japanese we need to have a prior word segmentation anyway in order to have Studio do something that makes sense, so I don't think it is an issue.

joanise:
> Yeah - ultimately what we want to do is to remove the spaces introduced by Unidecode but leave any others intact. For Chinese and (especially) Japanese we need to have a prior word segmentation anyway in order to have Studio do something that makes sense, so I don't think it is an issue.

Sounds good. I'll approve this then.

roedoejet (approved):
Thanks @dhdaines - looks good!

joanise:
@dhdaines We linearized the history as part of merging this in. I recommend you rebase your other branches onto `origin/main` (please rebase, please don't merge main back in) and force push the result onto your own fork, so that you keep working on top of that linear history.

PR #234, opened 2023-03-14
Misc improvements

joanise:
See the commit messages.


joanise:
Adding b, m, p, u rules to the moh mapping was requested by A.M. to make it explicit that these letters can happen in borrowed words. He provided the test words.

roedoejet (approved):
I think we should have a better way of handling this in the long run, but it makes sense for now. Thanks @joanise !

PR #240, opened 2023-03-30
refactor(test): when a mapping fails, show test case filename:lineno

joanise:
Bonus: in VSCode, you can click on that in the logs and it takes you to the test case file directly, from within tests/public/data.

I forgot to PR this change earlier.


dhdaines (approved):
seems like a good idea to me!

joanise:
What the heck? I'll have to check what's going on with coverage, this PR should have had 0 impact on it!

joanise:
Oh, wait, maybe it's because we're now combining coverage from two separate suites, and the first report I was looking at came too early and only saw one of the two.

Issue #241, opened 2023-04-12
g2p causing errors with pytest


roedoejet:
To reproduce (from Bronson Chapman):

```python
# test_calc.py
from g2p import make_g2p


def test_transducer():
    transducer = make_g2p('dan', 'eng-arpabet')
    output = transducer('hej').output_string

    assert output == 'HH EH Y'


def test_subtraction():
    assert 4 - 2 == 2
```

Then run

```bash
pip install pytest
pip install g2p
pytest
```

joanise:
Smaller example with the same output:
```python
# test_import.py
from g2p import make_g2p
```

We don't even need any tests in there, just the import of anything in g2p in a file called `test_whatever.py` does it.

Problem reproduced with Python 3.6, 3.8 and 3.10, so it's not version specific.

joanise:
Proposed patch that I'll test shortly:
```
-if sys.stdout.encoding != "utf8" and hasattr(sys.stdout, "buffer"):
+if "pytest" not in sys.modules and sys.stdout.encoding != "utf8" and hasattr(sys.stdout, "buffer"):
     sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding="utf8")

-if sys.stderr.encoding != "utf8" and hasattr(sys.stderr, "buffer"):
+if "pytest" not in sys.modules and sys.stderr.encoding != "utf8" and hasattr(sys.stderr, "buffer"):
     sys.stderr = io.TextIOWrapper(sys.stderr.buffer, encoding="utf8")
```

joanise:
That code is required for g2p to work correctly on a Windows terminal where the default encoding is not always (or even usually) utf8, but it's incompatible with pytest.

PR #242, opened 2023-04-12
fix: disable the utf8 fix for windows when running in pytest

joanise:
Minimal fix for #241.

This whole solution for making utf8 the default on Windows is not elegant in the first, and now I realize even in the readalongs CLI, we rely on `import g2p` to fix the stdout and stderr encoding, which is really not elegant!

This patch should make Bronson's project be able to proceed, but we should review how we make g2p and readalongs handle utf-8 on Windows.


roedoejet (approved):
As you mention, it's fairly hacky right now but I don't have a  great alternative for pytest and windows compat. I approve, and we can think about this later. Thanks @joanise !

PR #245, opened 2023-04-13
Make cli.py set its output encoding to utf8, and stop letting g2p/__init__.py do it.

### LINKED ISSUE #243: Clean up how we make g2p and readalongs handle utf8 by default on Windows
Currently, the g2p and readalongs CLI as well as anything else that imports g2p, for that matter, is (potentially unintentionally) running this block of code in `g2p/__init__.py` which resets the output to utf8 globally:

    if sys.stdout.encoding != "utf8" and hasattr(sys.stdout, "buffer"):
        sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding="utf8")
    if sys.stderr.encoding != "utf8" and hasattr(sys.stderr, "buffer"):
        sys.stderr = io.TextIOWrapper(sys.stderr.buffer, encoding="utf8")

That code does *not* belong in such a global location!!! (My bad, I wrote it...)

Instead, `g2p/cli.py` could do this after all its imports, and similarly `readalongs/cli.py`.

This was noticed when it interfered with using pytest, as documented in #241

joanise:
Until now, we relied on the side effect of `g2p/__init__.py` setting the stdout and stderr encoding to utf8, except that that was incompatible with pytest (see #241), and it really is not a good global practice at all. Any program driver that needs to set the output encoding should do it itself, it should not be a hidden side effect of importing a library.

Fixes #243

https://github.com/ReadAlongs/Studio/pull/176 should be merged before this PR is merged.


.github/workflows/tests.yml
```diff
@@ -52,3 +52,8 @@ jobs:
           pip install -e .
       - name: Run tests on Windows
         run: python run_tests.py dev
+      - name: Make sure the CLI outputs utf8 on Windows
+        run: |
+          # Warning: This is PowerShell syntax, not bash!
```
roedoejet, line 57: nice! didn't really think to do a test like this here, but it makes sense.

roedoejet (approved):
yep, this is better. thanks @joanise

PR #248, opened 2023-04-14
refactor: move get_langs from Studio/readalongs to g2p

joanise:
get_langs() is now use in both Studio/readalongs and everyvoice, so it belongs upstream in g2p itself, with suitable documentation as to what its output is.

Once this is merged and published, everyvoice and readalongs will be able to get DRYer.

In Studio, though (see branch `dev.factor-out-get_langs` in Studio), we still want to defer loading the g2p database until it's really needed, so we still have a `get_langs()` function that just imports and calls g2p's `get_langs()` function. everyvoice might not need to do the same thing, since it always imports g2p anyway.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ docs/_build/
 
 # coverage annotation output
 *,cover
+
+# vim temp files
+*~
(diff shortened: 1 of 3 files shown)


roedoejet (approved):
Nice multi-package refactoring here @joanise !

PR #257, opened 2023-05-11
fix(mappings): output should not be escaped

roedoejet:
We were escaping the "out" part of the mapping too, but that never gets compiled into a regular expression, so it was erroneously adding backslashes. Thanks Clarissa and Bronson!


dhdaines (approved):
Good catch! I think I may have noticed this at some point before, but didn't fix it :)

g2p/mappings/utils.py
```diff
@@ -430,10 +430,10 @@ def validate(mapping, path):
...
-            escaped = re.escape(v)
-        else:
-            escaped = v
+        if k == 'out':
+            to_escape[k] = v
+            continue
+        escaped = re.escape(v) if isinstance(v, str) else v
         if escaped != v:
             LOGGER.debug(
                 f"Escaped special characters in '{v}' with '{escaped}''. Set 'escape_special' "
```
joanise, line 439: superfluous second single quote after `'{escape}''`

joanise (approved):
Looks good!

PR #258, opened 2023-05-16
perf: optimize lexicon in-memory footprint

joanise:
Metric: how much RAM this command uses in ReadAlongs Studio:

    DEVELOPMENT=1 uvicorn readalongs.web_api:web_api_app --reload

Results:

     - baseline: eng lexicon stored as Dict[str, Tuple]       RAM: 100.7 MB
       where the tuple value is the parsed alignment
     - optimization 1: change to Dict[str, str] where the     RAM:  72.5 MB
       str value is the raw alignment string
     - failed opt: use a sorted list of tuples                RAM:  75.4 MB
     - optimization 2: change to List[str] with k, v          RAM:  61.9 MB
       joined by "\0", and bisect_left to find elements
     - absolute limit: no lexicon at all (v1.0.20230228)      RAM:  42.9 MB

So the English lexicon required about 60MB RAM in the previous (already optimized!) implementation and now shrinks to about 20MB.

Motivation: this will let us make the ReadAlong Studio web_api backend more memory efficient on our Heroku deployment, where memory is limited to 512MB per dyno.

### DIFF
--- a/g2p/mappings/utils.py
+++ b/g2p/mappings/utils.py
@@ -469,7 +469,6 @@ def load_abbreviations_from_file(path):
 
 def get_alignment_input_string(alignment: str) -> str:
     """Parse one alignment of the format in *.aligned.txt and return just the input"""
-    chars = ""
     return "".join(
         [
             tok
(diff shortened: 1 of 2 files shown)


joanise:
Hum, I better fix CI before going further with this!

joanise:
Argh! You're kidding me! The `key` parameter to `bisect_left()` was only added in Python 3.10!

g2p/mappings/utils.py
```diff
@@ -466,7 +467,81 @@ def load_abbreviations_from_file(path):
...
+        out_seq = delimiter.join(
+            tok for tok in mapping[idx + 1 :].split("|") if tok != "_"
+        )
+        # To save space, make the mappings flat and only store
+        # the number of input characters rather than the characters themselves
+        mappings.extend((in_len, out_seq))
+    return tuple(mappings)
+
+
+def parse_alignment(alignment: str, delimiter="") -> Tuple[str, Tuple]:
```
dhdaines, line 504: It seems that this function is not actually used anymore?

dhdaines (changes requested):
Good idea! Just a couple of small changes.

g2p/mappings/utils.py
joanise: Good point, it mildly nagged me, but not enough to delete it. I'll have to move its documentation where apporpriate, though.

joanise: Good point, it mildly nagged me, but not enough to delete it. I'll have to move its documentation where apporpriate, though.

g2p/mappings/utils.py
```diff
@@ -466,7 +467,81 @@ def load_abbreviations_from_file(path):
...
+    mappings: List[Union[int, str]] = []
+    for mapping in alignment.split():
+        idx = mapping.rindex("}")
+        # Note that we care about *character* indices, so we join them together
+        in_seq = "".join(tok for tok in mapping[:idx].split("|") if tok != "_")
+        out_seq = delimiter.join(
+            tok for tok in mapping[idx + 1 :].split("|") if tok != "_"
+        )
+        chars += in_seq
+        # To save space, make the mappings flat and only store
```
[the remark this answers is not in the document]
dhdaines: thanks! yes, it shouldn't require anything else.

joanise:
@dhdaines Good suggestions, the code is better now, thanks for the review! It's ready to review again.

dhdaines (approved):

PR #259, opened 2023-06-16
Adjust dependencies based on the Python version

joanise:
Python 3.6 is now still supported, but only for the CLI and the g2p library module, but not for g2p Studio.

Dependencies can now be updated as needed for Py >= 3.7 when there are CVEs.

### DIFF
--- a/.github/workflows/matrix-tests.yml
+++ b/.github/workflows/matrix-tests.yml
@@ -1,5 +1,7 @@
 name: Run full matrix Tests before Releases
 on:
+  push:
+    branches: [ main ]
   pull_request:
     branches: [ release ]
 jobs:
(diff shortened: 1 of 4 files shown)


joanise:
Note that I ran the full matrix test suite on this branch since I'm changing dependencies for different versions of Python.
See https://github.com/roedoejet/g2p/actions/runs/5292184153

joanise:
There is one known weakness here: I added a version test requiring python>=3.7 in `run_studio.py` but the normal way to launch the studio is via gunicorn, not via this script. However, if I move the test to `app.py` unit tests fail so I decided the test could just stay in that script.

roedoejet (approved):
Thanks @joanise !

PR #265, opened 2023-06-27
ci: reorganize CI test suites

joanise:
Before, the main tests were with Python 3.6, with only studio exercised with Python 3.8. This meant two uploads to codecov and an e-mail saying there was a failure (reduces coverage) systematically sent for each PR, because of the delay before the second upload happened.

I propose to have the main test suite run everything with Python 3.8, including studio, and do just one upload to codecov.

The 3.6 tests that are still relevant are only those making sure g2p still works as a library with 3.6, since we decided g2p-studio now requires >=3.7.


roedoejet (approved):
Nicely spotted and fixed. Thanks @joanise !

PR #269, opened 2023-07-06
feat: handle capitalization in kwk BOAS->Umista

joanise:
Draft PR. This is almost done, but the contextual rules don't work as intended.

When there was ambiguity (e.g., with small cap E and small cap L), I added `,,[A-Z]` to say make it upper case if the next letter is upper case, but default to the lower case rule otherwise.

There's lots going on here, first apply-longest-first sorting the files, and prevent-feeding swallowing context if the next character is processed first, but I've played around with several variants and I can't seem to get a handle on it. So help please!


joanise:
Doh, I just did things backwards: among rules of the same length, they are executed in the order they were found in the file, and I put the context sensitive rules after, not first.

joanise:
Figured it out:
 - put the context sensitive rules first, even when you use `apply-longest-first` - I had misconceptions about how ordering and apply longest first interact
 - I had some ambiguous rules I forgot to delete. `ᴇă,a̱ʼa` and `ᴇă,A̱ʼa` should not both exist since nothing distinguishes them, in contrast with `Ea` and `ea` being distinct inputs and therefore `Ea,E'a` and `ea,e'a` can and should both exist.

roedoejet (commented):
Do we really have to list them all extensionally like this? I think it would be worthwhile to have a preserve_case option that we can apply, particularly for non-IPA mappings, this would be a code change (and would break in some cases where capitalization doesn't adhere to Unicode specs), but all these rules make the mapping hard if not impossible to understand. I'm neither approving nor requesting specific changes, but my vote would be to not approve this as is, or at least to approve it temporarily in combination with an issue that sets us on a path of implementing a more general solution.

joanise:
I agree, it would be better to do this programmatically, rather than like this. 

Pat and I discussed that, he suggested the same as you, but I thought we don't actually have other use cases (at least not yet) and decided it didn't make sense to invest the time. The thing is, most g2p is to IPA, in which case case-preserving makes no sense, and we only have a few spelling-to-spelling mappings. That's why I didn't want to do it now.

But... you and Pat are both right, it would be much cleaner to have a case-preserving mode with rules like:
 - ALL CAPS -> ALL CAPS
 - all lower -> all lower
 - Capitalized -> Capitalized (ideally over tokens, rather than as individual characters are mapped)
 - MiXeD cApS -> ??? maybe in this case applied rule-by-rule in the mapping?

Special cases to handle cleanly:
 - small caps like `ɴʜʟᴇɶ` need to be spotted as "no case information" and considered not to break all caps/all lower/capitalized when the rest of the token follows that
 - ditto any other character that does not have casing variants, like `ɘ`, which suggests that the right rule is to test whether a character is case-invariant or not. That information is in the unicode data and easily programmatically available.

I like the name `preserve_case` for that option, which should be boolean and default to false, and should, when true, either require `case_sensitive: true` or else cause `case_sensitive` to be ignored (with a warning when case_sensitive is false).

littell (approved):
Approved, but we should probably still look into other possible solutions, like a preserve_case flag.

For rules like `ʟ,TŁ,,[A-Z]`, there are some Kwak'wala uppercase letters that don't fall into this range, including Ł and I think some of the combined forms of capitals with diacritics.  I'm not sure whether `re` has a way to just ask for any Unicode capital letter, but `regex` does, it's `\p{Lu}`.

joanise:
> For rules like `ʟ,TŁ,,[A-Z]`, there are some Kwak'wala uppercase letters that don't fall into this range, including Ł and I think some of the combined forms of capitals with diacritics. I'm not sure whether `re` has a way to just ask for any Unicode capital letter, but `regex` does, it's `\p{Lu}`.

Hum, yeah, good point. And the mapping is not normalizing to either NFC or NFD (this is probably a more general problem with this mapping, to be honest, with all the diacritics it uses) so we can't rely on the fact that the letter comes before the diacritic, which is actually pretty much what I'm doing here. It works with the test data, but it's not likely to be generally safe.

Pat, do you have a sense of whether this mapping really should be NFC normalized? Or maybe NFD? In fact, none of the kwk mappings set any normalization, I don't really understand why this has not been a problem in the past.

Looking at the rules, `kwk_boas_to_umista.csv` is itself NFD. If I convert it to NFC, some 90% of the rules change! Any NFC input fed to these mappings will cause trouble and basically appear to not work, at least not correctly.

Anyway, if we continue to assume NFD, `[A-ZŁ]` should capture all cases where the subsequent character is uppercase, but I'll look into whether we have access to something like `\p{Lu}`, that would be a far better solution.

joanise:
Hum, I was going to try to add `Ł` as a possible uppercase letter, as Pat mentioned, but I'm stumbling on all sorts of other bugs I didn't handle. I think my solution is completely broken anyway: my test case only passed because I was using `T` as the next letter needing to be identified as upper case context, but with any letter that actually gets mapped, prevent-feeding chews up my context after and things don't actually work as intended.
So, TL;DR, the only correct solution for kwk BOAS->Umista to handle capitalization correctly is indeed to implement `preserve_case`.

joanise:
We decided to abandon this approach and instead implement `preserve_case` to support this functionality. So this PR is replaced by #303

Issue #270, opened 2023-07-17
Add "preserve_case" options to g2p mappings


joanise:
#269 implements case preservation for kwk BOAS -> Umista by listing all the case possibilities. This is ugly!

Instead, we should have an option `preserve_case` (Boolean, default: false) which, when true:
 - causes `case_sensitive` to be ignored (or to issue a warning/error when it is false, or even just explicitly set?)
 - the input should be lower-cased, mapped with the mapping's rules following the current `case_sensitive: false` logic, and then
 - we apply these rules, token by token, by looking at the case of the original input:
   - ALLCAPS -> ALLCAPS
   - alllower -> alllower
   - Capitalized -> Capitalized (ideally over tokens, rather than as individual characters are mapped)
   - MiXeDcApS -> ??? maybe in this case applied rule-by-rule in the mapping?
 - Special cases to handle cleanly:
   - small caps like `ɴʜʟᴇɶ` need to be spotted as "no case information" and considered not to break all caps/all lower/capitalized when the rest of the token follows that
   - ditto any other character that does not have casing variants, like `ɘ`, which suggests that the right rule is to test whether a character is case-invariant or not. That information is in the unicode data and easily programmatically available.

roedoejet:
@joanise should this be closed since #274 was merged?

joanise:
Yes, we can close this, finally!

PR #274, opened 2023-07-18
Dev.casing

roedoejet:
Here's an attempt to address https://github.com/roedoejet/g2p/issues/270

I don't really understand the `kwk` mapping, so I just tried to address it in a general case. Curious to discuss with @joanise to see if it works sufficiently for what's needed for `kwk`.


g2p/transducer/__init__.py
```diff
@@ -1249,3 +1254,53 @@ def check(self, tg: TransductionGraph, shallow=False, display_warnings=False):
...
                     return False
         return result
+
+
+def preserve_case(
+    tg: TransductionGraph, case_equivalencies: Dict[str, str] = None
+) -> TransductionGraph:
+    if case_equivalencies is None:
+        case_equivalencies = {}
+    reverse_case_equivalencies = {v: k for k, v in case_equivalencies.items()}
```
joanise, line 1264: I guess it's OK to assume this is a 1-1 mapping, as you do here.

g2p/tests/test_transducer.py
```diff
@@ -209,6 +210,42 @@ def test_deletion(self):
...
+        transducer = Transducer(mapping)
+        self.assertEqual(transducer("'a").output_string, "b")
+        self.assertEqual(transducer("'A").output_string, "B")
+        self.assertEqual(transducer("E\u0301").output_string, "F")
+        self.assertEqual(transducer("E\u0301").output_string, "F")
+        # Test what happens in Heiltsuk. \u03BB should be capitalized as \u2144
+        self.assertEqual(transducer("TLaba").output_string, "\u2144aba")
+        self.assertEqual(transducer("tlaba").output_string, "λaba")
+        # I guess it's arguable what should happen here, but I'll just change case if any of the characters are differently cased
+        self.assertEqual(transducer("Tlaba").output_string, "\u2144aba")
```
joanise, line 234: What do we do in the opposite scenario? kwk has complexities that I'd like to know are addressed. E.g., I think this is what we want:

context based:

    ᴇTO -> ETO
    ᴇto -> eto

1 char -> 2 char context based:

    ʟat -> tłat
    ʟAT -> TŁAT

2 char -> 3 char context based:

    EᴇT -> EʼA̱T
    Eᴇt -> E'a̱t

and I'll want to test what preserve_case does in all these cases when we turn it on for kwk BOAS->Umista.

g2p/transducer/__init__.py
```diff
@@ -1249,3 +1254,53 @@ def check(self, tg: TransductionGraph, shallow=False, display_warnings=False):
...
+    for item in tg.substring_alignments():
+        in_sub = item[0]
+        out_sub = item[1]
+        any_in_upper = any(x.isupper() for x in in_sub)
+        any_in_lower = any(x.islower() for x in in_sub)
+        any_out_upper = any(x.isupper() for x in out_sub)
+        any_out_lower = any(x.islower() for x in out_sub)
+        # continue if character is un-caseable
+        if (
+            out_sub not in case_equivalencies
```
joanise, line 1277: So you're assuming case_equivalencies only lists complete rule outputs, rather than characters? What about `ʟ -> tλ` which would rely on the standard "t":"T" but would have to declare "λ":"⅄"?

g2p/transducer/__init__.py
```diff
@@ -1249,3 +1254,53 @@ def check(self, tg: TransductionGraph, shallow=False, display_warnings=False):
...
+            new_string += out_sub.upper()
+            continue
+        # change to lower if required
+        if any_in_lower and any_out_upper:
+            new_string += out_sub.lower()
+            continue
+        # just in case, append the out_sub
+        new_string += out_sub
+    tg.output_string = new_string
+    return tg
```
joanise, line 1306: Boy, it's hard to follow all this logic! We need more extensive unit testing just to make sure it works.

E.g., exercise these:
 - Given rule `aa -> bb`, `Aaxyz` should get mapped to `Bbxyz`
 - the hard stuff in kwk BOAS with those small caps being mapped to a context-dependent case
 - Given rule `a -> bb` `AXYZ` should get mapped to `BBXYZ` but `Axyz` to `Bbxyz` (although this is actually not fully reliable, e.g., in Spanish in some cases `ll` is capitalized as a unit, e.g. `llama` might capitalize to `LLama` though from my experience that's not a uniformly applied rule. And does kwk Umista consider `tł` as a single letter capitalized jointly, or as two letters with only the T being uppercased in capitalized words?)

g2p/tests/test_transducer.py
```diff
@@ -209,6 +210,42 @@ def test_deletion(self):
...
+            ],
+            case_sensitive=False,
+            preserve_case=True,
+            norm_form="NFC",
+            case_equivalencies={"λ": "\u2144"},
+        )
+        transducer = Transducer(mapping)
+        self.assertEqual(transducer("'a").output_string, "b")
+        self.assertEqual(transducer("'A").output_string, "B")
+        self.assertEqual(transducer("E\u0301").output_string, "F")
```
joanise, line 228: You probably meant to lowercase one of these two lines
```suggestion
        self.assertEqual(transducer("e\u0301").output_string, "f")
```

g2p/tests/test_transducer.py
```diff
@@ -209,6 +210,42 @@ def test_deletion(self):
...
+        self.assertEqual(transducer("E\u0301").output_string, "F")
+        # Test what happens in Heiltsuk. \u03BB should be capitalized as \u2144
```
joanise, line 230: We can help the reader of the code by inserting the rendered characters
```suggestion
        # Test what happens in Heiltsuk. \u03BB (λ) should be capitalized as \u2144 (⅄)
```

joanise (commented):
I'm not sure this handles all the hard cases correctly. Leaving a bunch of comments now, then I'll want to apply preserve case to kwk BOAS->Umista and test to see what happens with my unit test cases for that, and then I'll probably have more feedback.

At the very least this code needs quite a bit more unit testing to capture more tricky cases. See inline comments.

g2p/transducer/__init__.py
roedoejet: yes because we don't really have a notion of characters anywhere (re: the tokenization issue from before). here you would have to do `{'tλ': 'T⅄'}` I suppose

g2p/transducer/__init__.py
roedoejet: > Boy, it's hard to follow all this logic! We need more extensive unit testing just to make sure it works.
> 
> E.g., exercise these:
> 
> * Given rule `aa -> bb`, `Aaxyz` should get mapped to `Bbxyz`

Yes, this works.

> * the hard stuff in kwk BOAS with those small caps being mapped to a context-dependent case

I don't fundamentally really understand what is going on here, and I can't reproduce the examples you have above in the rules even without changing anything so we need to talk more about this.

> * Given rule `a -> bb` `AXYZ` should get mapped to `BBXYZ` but `Axyz` to `Bbxyz` (although this is actually not fully reliable, e.g., in Spanish in some cases `ll` is capitalized as a unit, e.g. `llama` might capitalize to `LLama` though from my experience that's not a uniformly applied rule. And does kwk Umista consider `tł` as a single letter capitalized jointly, or as two letters with only the T being uppercased in capitalized words?)

I disagree that for a rule a -> bb Axyz should get turned to Bbxyz. We will handle this as BBxyz, otherwise someone would need to add a case equivalency for bb and Bb, in which case it should work.

joanise: I guess we might need to ask Daisy what she thinks is right for kwk. And I guess we should also test with kwk and see what comes out with your current implementation.

So for kwk my assumption has been that you can infer whether a small cap L or a small cap E should be interpreted as upper case by the case of the next letter. I suppose we should validate that assumption with Daisy before investing too much development effort making it work...

joanise:
I'm playing with case_equivalencies now, to enhance the kwk mapping.
For example, for rule `ᴇa,a̱ʼa`, I'd like to declare the equivalency `"a̱ʼa": "A̱ʼA", so that `TᴇAtᴇa` could get mapped to `TA̱ʼAta̱ʼa` rather than `Ta̱ʼAta̱ʼa` but I can't figure out how to make it happen.
A challenge, here, is that for this kwk mapping we'll have to have declare a couple dozen equivalencies to make all the vowel pairs that include `ᴇ` work correctly.

joanise:
@roedoejet 
Rebased and improved. While still nor perfect, I think it's time to merge this as good enough, assuming CI passes. It will help most of the time, and the errors in case preservation for kwk will just mean once in a while, a character that should have been case-preserved is not. I think it's better to accept this partial solution than do the perfect solution I was hoping for.

But boy, rebasing this over the pydantic update was not trivial! Please review my work carefully.

joanise:
Hum, I can't request a review from you @roedoejet because it was your PR in the first place. Review it anyway and comment here.

g2p/tests/test_mappings.py
```diff
@@ -187,6 +187,10 @@ def test_case_sensitive(self):
         self.assertEqual(transducer_case_sensitive("a").output_string, "a")
         self.assertEqual(transducer("A").output_string, "b")
 
+    def test_case_equivalencies(self):
+        with self.assertRaises(exceptions.MalformedMapping):
```
roedoejet, line 191: this is a good compromise

g2p/tests/test_transducer.py
```diff
@@ -218,6 +219,42 @@ def test_deletion(self):
...
+        )
+        transducer = Transducer(mapping)
+        self.assertEqual(transducer("'a").output_string, "b")
+        self.assertEqual(transducer("'A").output_string, "B")
+        self.assertEqual(transducer("E\u0301").output_string, "F")
+        self.assertEqual(transducer("e\u0301").output_string, "f")
+        # Test what happens in Heiltsuk. \u03BB (λ) should be capitalized as \u2144 (⅄)
+        self.assertEqual(transducer("TLaba").output_string, "\u2144aba")
+        self.assertEqual(transducer("tlaba").output_string, "λaba")
+        # I guess it's arguable what should happen here, but I'll just change case if any of the characters are differently cased
```
roedoejet, line 242: I think these tests are pretty clear with what our implementation is

g2p/mappings/utils.py
```diff
@@ -661,7 +661,13 @@ class _MappingModelDefinition(BaseModel):
     """Deprecated: Please use rule_ordering='as_written' """
 
     case_sensitive: bool = True
```
roedoejet, line 663: I would like to see some tests about interactions between `case_sensitive` and `preserve_case`. Does `case_sensitive=False` and `preserve_case=True` work for example?

roedoejet (commented):
Looks good to me - I can't approve it, since it's my own PR. My only remaining question is the one about interactions between `case_sensitive` and `preserve_case`

g2p/mappings/utils.py
joanise: Hum, good question. It actually only makes sense to have `case_sensitive=False` when you set `preserve_case=True`. Having both be true should raise an error.

joanise: The semantics of preserve case is that case does affect the meaning of the characters (i.e., the mapping is case insensitive) but should be preserved because it has non-phonetic meaning (e.g., proper name, beginning of the sentence).

joanise: So I just added a model validator that raises when the two are set incompatibly.

joanise (approved):

Issue #276, opened 2023-08-04
Strange error with moh mapping


roedoejet:
This rule eats the glottal stop :(

  `{   'end': 7,
                    'input': "sɑɡo'shʌ̃́nje's",
                    'output': "sɑɡosʌ̃́nje's",
                    'rule': {   'context_after': 'a|á:|à:|a:|à|á|ɑ|ɑ́|ɑ̀|ɑ̀ː|ɑ́ː|ɑː|ʌ̃|ʌ̃́ː|ʌ̃ː|ʌ̃̀ː|ʌ̃̀|ʌ̃́|én:|en|en:|èn:|én|èn|e|è:|é|é:|eː|éː|èː|i|i:|í:|í|ì:|ì|iː|íː|ìː|ón:|òn:|on:|òn|on|ón|ṹː|ũ̀ː|ũ̀|ũ|ṹ|ũː|o|ó|ó:|o:|ò:|ò|óː|oː|òː',
                                'context_before': "h|k|n|n̥|r|ɽ|ɽ̥|t|s|w|'|t͡s|d͡ʒ|d͡z|t͡ʃ|ʃ|ʒ|d|ɡ|ɡʷ|kʷ|kʰʷ|f|j|z|ʔ",
                                'in': 'sh',
                                'out': 's'},`

roedoejet:
and here :( 

`{   'end': 3,
                    'input': "o'nhónhsɑ'",
                    'output': "on̥hónhsɑ'",
                    'rule': {   'context_after': "h|k|n|n̥|r|ɽ|ɽ̥|t|s|w|'|t͡s|d͡ʒ|d͡z|t͡ʃ|ʃ|ʒ|d|ɡ|ɡʷ|kʷ|kʰʷ|f|j|z|ʔ",
                                'context_before': "h|k|n|n̥|r|ɽ|ɽ̥|t|s|w|'|t͡s|d͡ʒ|d͡z|t͡ʃ|ʃ|ʒ|d|ɡ|ɡʷ|kʷ|kʰʷ|f|j|z|ʔ",
                                'in': 'n',
                                'out': 'n̥'},`

Issue #275, opened 2023-08-04
Error combining abbreviation and regular expression special character


roedoejet:
This rule:

`{ "in": "s", "out": "z", "context_after": "VOWEL", "context_before": "VOWEL|^" }`

Works for "seb" -> "zeb", but not "eseb" -> "ezeb" (still produces "eseb")

Issue #268, opened 2023-07-06
"Add Rule" doesn't work in g2p studio


roedoejet:
When selecting a language other than custom, "Add rule" does nothing

Issue #208, opened 2022-12-16
use the endonym for moh language name?


joanise:
For `moh`, we should probably update the config to use the endonym `Kanyen'kehà:ka` as the language name, instead of Mohawk. Or maybe `Kanyen'kehà:ka (Mohawk)`?

roedoejet:
Yes, although Kanyen'kehà:ka is the name of the people, not language. It should be Kanyen'kéha or Kanien'kéha depending on dialect. The problem with endonyms is there are dialect differences that we would either need to choose from or list extensionally. I agree this would be the right thing to do though.

dhdaines:
Yeah, though of course if different dialects have different orthographies then they are de facto different "languages" for our purposes!

PR #277, opened 2023-08-15
A variety of bug fixes from working with AM

roedoejet:
There were a series of small related bug fixes with the creation of fixed-width lookbehinds. This PR:

- adds Akwiratékha' Martin's updated moh -> moh-ipa mapping (along with updated abbreviations and unit tests) with some minor updates and re-orderings
- removes the defunct moh-festival mapping
- fixes a bug in the haa mapping
- adds :'^$ to the fixed width lookbehind pattern match - there should ideally be more matched here. (fixes https://github.com/roedoejet/g2p/issues/276)
- fixes a bug where $ and ^ are null/0-length and so they need to be separate from other single-character lookbehind patterns (fixes https://github.com/roedoejet/g2p/issues/275) 
- changes name of language (fixes https://github.com/roedoejet/g2p/issues/208)
- Fix 'add rule' by pinning version of hands on table (fixes https://github.com/roedoejet/g2p/issues/268)


roedoejet:
Does this help fix https://github.com/roedoejet/g2p/issues/134 @joanise ?

joanise (approved):
Lots of stuff bundled together here, but I've reviewed it commit by commit and it all looks good. I see that changes are well tested. I didn't take the time to fully understand each change, but it all seems good. Thanks for the links to the issues being fixed, in the commit messages, that made it much easier to follow changes along.

About my comment concerning the extra spaces in moh_to_ipa, it would be nice to remove them to keep this file closer to what we do elsewhere in langs/*.json, but it's not a required change, so selecting "approve" rather than "request changes".

PR #279, opened 2023-08-25
fix(crg): fix various rule feeding and ordering bugs for Michif

joanise:
Both Michif -> IPA mappings had issues due to incorrect rule ordering which mean that some words, e.g. "booñ" or "not", would not get g2p'd correctly.

as-written is still required because apply-longest-first would break the "g" rule, which does indeed need to be applied first for both DV and TMD. But a rule whose input is a superstring of another rule must come first, and that was not respected everywhere.

Note, in TMD I know "n,,in|an|en|un|on|ɑːn|æn|ɛn|eːn|ɪn|iːn|oːn|uːn" is super ugly. I was trying to express context_before is VOWEL + "n", but I don't know how to do it. None of these work: [VOWEL]n (VOWEL)n VOWELn


dhdaines (approved):
I rebased this with the latest main branch.  It looks good, and is definitely more correct than what was there before.  I am going to merge it then rebase P2WILRC's branch on top of that - I won't do a PR for that branch quite yet though as it mostly just adds TMD to DV conversion which isn't entirely functional.

PR #284, opened 2023-09-06
refactor(docs): change docs to mkdocs

roedoejet:
This PR has:

- updated the docs to be served on gh-pages instead of readthedocs
- added versioning to the docs such that new pushes update `latest` and version releases are pinned to specific versions as well as `stable` and `latest`
- added comments sections to the documentation that is linked with our `Discussions` page via giscus


.github/workflows/tests.yml
```diff
@@ -1,7 +1,9 @@
 name: Run all tests
 on:
-  - push
-  - pull_request
+  pull_request:
+      types: [review_requested]
+  push:
+
```
joanise, line 6: Will this still run when a PR is submitted from a fork? I'm guessing that's what the `review_requested` thing tries to do, but what happens when a PR is on a fork, review is requested, then a commit is added on that fork, will the tests run again?

In any case, I like the idea of removing the duplicate run, it's redundant the way we run both on the push and the PR, so I'm glad you're looking into a way to optimize that redundancy out.

roedoejet: > Will this still run when a PR is submitted from a fork? I'm guessing that's what the review_requested thing tries to do, but what happens when a PR is on a fork, review is requested, then a commit is added on that fork, will the tests run again?

I am not totally sure. I believe it would though since it's triggered on each push as well. So this way a PR that gets made without any pushes won't get triggered until we request review, but subsequent commits should still trigger the tests

joanise: I think a push on a fork is not a push on the parent repo, so I don't know if it would get triggered.

I guess the thing to do would be to test it in a dummy fork.

README.md
```diff
@@ -114,6 +144,21 @@ JSON files are written as an array of objects where each object corresponds to a
  ]
 ```
 
+The equivalent python syntax would be:
```
joanise, line 147: Instead of repeating this under each CSV example and again under each JSON example, let's make this a three subsection things, with `#### CSV` followed by `#### JSON` and finally `#### Python`.

README.md
```diff
@@ -175,7 +235,7 @@ So, when you write a new g2p mapping for a language, say `lll`, and you want to
 1. Write the mapping from `lll` to `lll-ipa` in g2p/mappings/langs/lll/. You've just updated Text DB.
 2. Run `g2p update` to regenerate Compiled DB from the current Text DB and Gen DB, i.e., to incorporate your new mapping rules.
 3. Run `g2p generate-mapping --ipa lll` to generate g2p/mappings/langs/generated/lll-ipa\_to\_eng-ipa.json. This is not based on what you wrote directly, but rather on what's in Generated DB.
-4. Run `g2p udpate` again. `g2p generate-mapping` updates Gen DB only, so what gets written there will only be reflected in Compiled DB when you run `g2p update` once more.
+4. Run `g2p update` again. `g2p generate-mapping` updates Gen DB only, so what gets written there will only be reflected in Compiled DB when you run `g2p update` once more.
```
joanise, line 238: Wow, this typo is still here! It feels to me like I've corrected it multiple times, but maybe not, or maybe elsewhere... :)

joanise: It's a typo I make all the time, I can never type `g2p update` without first typing `g2p udpate`... And I introduced it into the code base here... :)

docs/index.md
```diff
@@ -0,0 +1,5 @@
+# Welcome to the G2P documentation!
+
+G2P is a tool for doing rule-based conversions for text.
+
+This website has the technical documentation for G2P, but we've also written a [7-part blog series](https://blog.mothertongues.org/g2p-background/) if you want a more thorough introduction to why G2P exists and what you can use it for.
```
joanise, line 5: It feels very naked when you land on the home page and this is all you see, you have to scan to find that the navigation menu is on the left. I think I would add a TOC here under this paragraph, or at least put a note saying the navigation is on the left.

docs/package.md
```diff
@@ -0,0 +1,33 @@
...
+
+## A look under the hood
+
+A Mapping object is a list of defined rules. A `Rule` has the following permitted fields:
+
+::: g2p.mappings.Rule
+    options:
+        show_root_heading: true
+        show_source: false
+        heading_level: 4
```
joanise, line 33: Why did you remove Transducer docs from here?

joanise (changes requested):
Is https://roedoejet.github.io/g2p/latest/ currently pointing to what this generates? If so, I'd like to add Next/Previous buttons on each page, and I'm not sure I'd keep the comment section on each page.
Overall this seems good. See further comments inline.

joanise (approved):

PR #286, opened 2023-09-11
final patches on pydantic PR

joanise:
Submitted as a new PR on top of yours because I want your feedback on removing the four dunder methods on Mapping before actually doing it.


roedoejet (approved):
thanks for fixing the 3.7 compat issue - about the dunder methods, I don't mind changing it, but it does introduce another breaking change. Now, there are already a lot of breaking changes in 2.0 so I suppose this is a good time to do it, but it also doesn't seem necessary. I'm indifferent though, so let's just accept this. Thanks @joanise !

PR #288, opened 2023-09-12
Use JSON for the network file

dhdaines:
- Switch to using compressed JSON for the network as well as the langs file
- Cease to use `JSON` and `PKL` in the constant names in `g2p.mappings.langs` in case we decide to change formats again
- Remove curiously specific `networkx` dependency


g2p/mappings/langs/utils.py
```diff
@@ -169,22 +169,23 @@ def cache_langs(
...
+    with gzip.GzipFile(path, "wb", mtime=0) as zipfile:
+        zipfile.write(
+            json.dumps(
+                data,
                 separators=(",", ":"),
                 ensure_ascii=False,
                 sort_keys=True,
-            )
-    return langs
+            ).encode("utf-8")
```
joanise, line 187: Wow, haven't seen the use of `encode()` in a long time! I kinda forget that exists, these days...

g2p/mappings/langs/utils.py
```diff
@@ -169,22 +169,23 @@ def cache_langs(
...
 
-    with open(network_path, "wb") as f:
-        write_gpickle(lang_network, f, protocol=4)
 
-    with gzip.GzipFile(langs_path, "wb", mtime=0) as zipfile_raw:
-        with io.TextIOWrapper(zipfile_raw, encoding="utf-8") as zipfile:  # type: ignore
-            json.dump(
-                langs,
-                zipfile,
+def write_json_gz(path: str, data: str):
```
joanise, line 179: OK, this was worth factoring out, I now agree. 😄

requirements/requirements.txt
```diff
@@ -6,7 +6,7 @@ eventlet>=0.33.0
 flask-cors>=3.0.9
 flask-talisman>=0.7.0
 flask-restful>=0.3.9
-networkx>=2.5,<=2.8.4
+networkx>=3.1
```
joanise: This will break on Python 3.7.

joanise: The most recent release of `networkx` for Python 3.7 is 2.6.3... ☹️

joanise: But the unit tests pass with `>=2.6.3` so we don't need to ask for 3.1

roedoejet (approved):
Looks great to me, thanks @dhdaines !!

joanise (changes requested):
I'm glad we have good unit tests. This is good. We just have to keep networkx compatible with Python 3.7 and we can merge this.

joanise:
And I should have hit that refresh button GitHub was displaying, you've already made the changes! 😄

joanise (approved):

PR #289, opened 2023-09-12
feat(schema): update schema generation to include  dialect spec by default

roedoejet:
SchemaStore wants the $schema key to be populated, but this is not done by default by pydantic. unfortunately pydantic also targets a different jsonschema dialect than the one recommended by SchemaStore - I'll check to see if this is acceptable.


joanise:
Looks like this was an easy thing to fix in the end. Or at least the solution looks simple.

PR #291, opened 2023-09-26
Fix failure in test failure

dhdaines:
If a test CLI fails, then the error message fails also, because we expect the wrong contents in a tuple (we did not notice this because no tests actually fail):
```
======================================================================
FAIL: test_convert (__main__.CliTest)
----------------------------------------------------------------------
  File "/home/dhd/work/g2p/g2p/tests/test_cli.py", line 148, in test_convert
    self.assertEqual(
AssertionError: "Usage: convert [OPTIONS] INPUT_TEXT IN_L[86 chars]ipa)" != 'bɛ̃ː'
+ bɛ̃ː- Usage: convert [OPTIONS] INPUT_TEXT IN_LANG OUT_LANG
- Try 'convert -h' for help.
- 
- Error: Got unexpected extra argument (crg-ipa) : crg-tmd->crg-ipa mapping error for '{first_failed_test[2]}'.
Look for warnings in the log for any more mapping errors

----------------------------------------------------------------------
Ran 1 test in 12.412s

FAILED (failures=1)
```
This changes the code to destructure said tuple to make it a bit more obvious, and also fixes the problem.

### DIFF
--- a/g2p/tests/test_cli.py
+++ b/g2p/tests/test_cli.py
@@ -121,35 +121,43 @@ class CliTest(TestCase):
         )
         error_count = 0
         for tok_option in [["--tok", "--check"], ["--no-tok"]]:
-            for test in langs_to_test:
+            for row in langs_to_test:
+                (in_lang, out_lang, word_to_convert, reference_string, fileline) = row[
+                    :5
+                ]
                 output_string = self.runner.invoke(
-                    convert, [*tok_option, test[2], test[0], test[1]]
+                    convert, [*tok_option, word_to_convert, in_lang, out_lang]
                 ).stdout.strip()
-                if test[3].strip() not in output_string:
+                if reference_string.strip() not in output_string:
                     LOGGER.warning(
-                        f"test_cli.py for {test[-1]}: {test[0]}->{test[1]} mapping error: '{test[2]}' "
-                        f"should map to '{test[3]}', got '{output_string}' (with {tok_option})."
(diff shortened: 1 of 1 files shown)


roedoejet (approved):
Ah yes, this is better to avoid IndexErrors - thanks @dhdaines !

g2p/tests/test_cli.py
```diff
@@ -121,35 +121,43 @@ def test_convert(self):
         )
         error_count = 0
         for tok_option in [["--tok", "--check"], ["--no-tok"]]:
-            for test in langs_to_test:
+            for row in langs_to_test:
+                (in_lang, out_lang, word_to_convert, reference_string, fileline) = row[
+                    :5
+                ]
```
joanise, line 127: There's a bug here. `fileline` is not `row[4]`, but `row[-1]`. Most of the time, that's the same thing, but our test data allow adding an extra column for comments. See `moh.psv` for example. That's why we were using `test[-1]` in the previous code rather than `test[4]`.

So this should probably be
```
(in_lang, out_lang, word_to_convert, reference_string, *_, fileline) = row
```

dhdaines: Ah, oops, you're right, I missed that.  I didn't realize the syntax above was possible either, in fact it is what I was looking for in this case but couldn't find (horror of horrors, I assumed that it would be `...` like in JavaScript!)

So another PR will arrive in a second.

PR #292, opened 2023-09-28
Fix fix to test faliure

dhdaines:
Turns out that [-1] was there for a reason, there can be arbitrary extra fields (but really just comments) in a test file.  Thanks @joanise  for pointing this out!

### DIFF
--- a/requirements/requirements.dev.txt
+++ b/requirements/requirements.dev.txt
@@ -2,6 +2,7 @@ black~=22.0
 flake8>=4.0.1
 gitlint-core==0.17.0
 isort>=5.10.1
-mypy>=0.941,<=0.971
+mypy>=1.4.1
+mypy>=1.5.1; python_version >= "3.8"
 pre-commit>=2.6.0
 types-PyYAML
--- a/g2p/mappings/langs/utils.py
+++ b/g2p/mappings/langs/utils.py
@@ -5,9 +5,9 @@ Utilities used by other classes
 """
 
 import gzip
-import io
 import json
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


requirements/requirements.dev.txt
```diff
@@ -2,6 +2,7 @@ black~=22.0
 flake8>=4.0.1
 gitlint-core==0.17.0
 isort>=5.10.1
-mypy>=0.941,<=0.971
+mypy>=1.4.1
```
roedoejet, line 5: would this not also possibly install 1.5.1 for 3.7? Is that what you want? This just sets a different min version for 3.8 and higher.

roedoejet (commented):
This looks good to me, but I have a comment/question for the requirements change

requirements/requirements.dev.txt
dhdaines: Hmm.  Not sure ... I just copied it from Studio.

dhdaines: I think @joanise was the one who introduced it there?  My guess is that 1.5.1 isn't actually installable on 3.7, and we don't specifically depend on its features, but we'd *like* to have it if possible elsewhere.

joanise: Exactly, the most recent mypy available for Python 3.7 is 1.4.1.

My commit log message for that change says "fix: make Studio compatible with Pydantic 2 and thus g2p 2", and I'm sure I had a good reason to do it this way, but that was a whole month ago, so I don't remember what it was.

PR #293, opened 2023-10-02
Produce valid `config-g2p.yaml` when running `g2p generate-mapping`

dhdaines:
The `config-g2p.yaml` produced by `g2p generate-mapping` was quite incorrect, as it included the `rules` in the configuration, and this, in an arbitrary order, thus rendering `rule_ordering: as-written` inoperative and the JSON files irrelevant.

This fixes that and also omits spurious fields, giving output that more or less resembles the previous version.  However, the names of some JSON files have also changed, so we have to move them to the new names.


dhdaines:
Unfortunately there seem to be various other problems with the generated mappings, so this can't be merged yet.

joanise:
For what it's worth, the rule ordering should only be changed if `apply-longest-first` is enabled, in which case the `rules:` should be listed by reverse order of length of input. If `as-written` were specified, I believe it would be a bug if the rules got re-ordered when printed by `export_to_dict()`.

g2p/mappings/langs/generated/config-g2p.yaml
```diff
@@ -1,472 +1,333 @@
...
-    norm_form: NFD
+    language_name: SENĆOŦEN
     out_lang: eng-ipa
     rule_ordering: apply-longest-first
+    rules_path: str-ipa_to_eng-ipa.json
   - author: Generated 2019-10-24 16:21:41.605592
     authors:
       - Generated 2022-04-12 11:26:30.098188
-    display_name: "SEN\u0106O\u0166EN IPA to English IPA"
+    display_name: SENĆOŦEN IPA to English IPA
```
joanise, line 39: Ah, much nicer!

g2p/mappings/__init__.py
```diff
@@ -318,18 +318,32 @@ def mapping_to_file(self, output_path: str = GEN_DIR, file_type: str = "json"):
...
+    def export_to_dict(self, mapping_type="json", config_only=False):
+        """Export a mapping to a dictionary, optionally including only the
+        configuration (and not the rules or alignments) in the case
+        where we are just writing the config file.
+        """
+        if config_only:
+            model_dict = self.model_dump(
+                mode="json",
+                exclude_none=True,
+                exclude_defaults=True,
```
joanise, line 330: Oh, I don't like this much. It resulted in all the `norm_form: NFD` declarations being removed. While equivalent for the machine, it's less clear for the human reader...

joanise: Ditto all the `rule_ordering: as-written`

joanise (changes requested):
This is an important patch, thanks!
I agree with `exclude={"parent_dir": True, "rules": True, "processed": True},` but can we remove `exclude_defaults=True,`?

g2p/mappings/__init__.py
dhdaines: Yes, makes sense, I mainly wanted it to remove the `abbreviations` and `alignments` keys, but actually we should just explicitly remove these when `config_only=True` because they should not exist in a configuration file either.  Done in the commit below!

dhdaines:
> For what it's worth, the rule ordering should only be changed if `apply-longest-first` is enabled, in which case the `rules:` should be listed by reverse order of length of input. If `as-written` were specified, I believe it would be a bug if the rules got re-ordered when printed by `export_to_dict()`.

Right, this is exactly what was happening, the rules were being re-ordered.  This was probably because of the dumping to JSON and reloading as a `dict` that we were doing.  *But* the rules shouldn't have been in `config-g2p.yaml` anyway...

joanise:
Argh, we just can't win, can we!?

Before Pydantic, `generated/config-g2p.yaml` had a stable number of fields, triggered by whether the field was in the file in the first place, rather than whether it was the same as the previous default or not. So re-running another generate mapping added one mapping and left everything else in the file intact.

Now, based on your initial version of this PR and this updated version, it looks like you can trigger include/exclude based on if the field still has its default value, which is not actually what I'd want!

Ideally I'd want the old behaviour, but now I'm not sure that's available to us. Let me do some research on that...

dhdaines:
> Now, based on your initial version of this PR and this updated version, it looks like you can trigger include/exclude based on if the field still has its default value, which is not actually what I'd want!

No the updated version writes *all* the fields except ones that shouldn't be there.  From what I understand `config-g2p.yaml` (anywhere, not just in `generated`) shouldn't have `rules`, `abbreviations`, or `alignments` fields.  So we just don't write them in this case.

dhdaines:
> Yeah, I'm looking for a way around this. I'm hoping it's still possible to have the old behaviour, but maybe it just isn't.

Yes, we can't have the old behaviour, where only the fields which are explicitly set (whether or not they were set to the default values) are included in the file.

joanise:
Then the new behaviour with the latest changes you added are the better option, so let's go with it.

joanise (approved):

PR #294, opened 2023-10-02
More crg-dv fixes

dhdaines:
This gets us through a good sampling of the southernmichif.org texts.  Surely there are some more vowels missing here and there.


joanise (approved):
I trust these changes reflects information from appropriate sources. Looks good to me as far as I can see. Thanks for adding unit test cases to support it.

joanise:
You can ignore the drop in coverage, this PR changes no code.

PR #298, opened 2023-11-08
fix: remove werkzeug lock since it is no longer necessary

joanise:
we had locked it to an older version when we were using a flask-socketio that was not compatible with the latest werkzeug, but we bumped that now so we no longer need this, and keeping it makes dependabot unhappy so remove it.

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -15,7 +15,6 @@ tqdm
 text_unidecode
 
 Flask==2.2.5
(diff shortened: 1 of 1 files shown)


roedoejet (approved):
lgtm

joanise:
CI installed werkzeug 3.0.1 with Python 3.8 and 2.2.3 with Python 3.7, this should quiet dependabot now.

Issue #299, opened 2023-11-09
the ordering for apply longest first includes the indices


joanise:
To reproduce, add this test case to `test_mappings.py`:
```
    def test_rule_ordering_with_indices(self):
        """a{1}b{3} should be shorter than abc"""
        rules = [{"in": "a{1}b{2}", "out": "x{1}x{2}"}, {"in": "abc", "out": "y"}]
        mapping = Mapping(rules=rules, rule_ordering="apply-longest-first")
        transducer = Transducer(mapping)
        self.assertEqual(transducer("abc").output_string, "y")
```
it will fail, saying:
```
FAIL: test_rule_ordering_with_indices (__main__.MappingTest)
a{1}b{3} should be shorter than abc
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./test_mappings.py", line 186, in test_rule_ordering_with_indices
    self.assertEqual(transducer("abc").output_string, "y")
AssertionError: 'xxc' != 'y'
- xxc
+ y
```

PR #300, opened 2023-11-09
fix: sort rules without explicit indices

roedoejet:
fixes: https://github.com/roedoejet/g2p/issues/299

### DIFF
--- a/g2p/mappings/__init__.py
+++ b/g2p/mappings/__init__.py
@@ -37,6 +37,7 @@ from g2p.mappings.utils import (
 )
 
 GEN_DIR = os.path.join(os.path.dirname(LANGS_FILE), "generated")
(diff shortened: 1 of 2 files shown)


joanise (approved):
Thanks for fixing this so quickly!

joanise:
I'm amazed this PR did not require re-running `g2p update`. I guess that's why we never noticed it before, it didn't make a difference. It will with my current work on kwk, though.

joanise:
> I'm amazed this PR did not require re-running `g2p update`. I guess that's why we never noticed it before, it didn't make a difference. It will with my current work on kwk, though.

Oh, it's because langs.pkl has the rules in the order they are written, not in the order they are applied. dum-dee-dum

joanise:
And yes, this patch will already change the order in which the rules in `umista-equiv.csv` are applied.

PR #301, opened 2023-11-15
refactor: move strip_index_notation to mappings/utils.py

joanise:
I think this new function is a great idea, but it belongs in the existing utils.py file.

Also, the doctest test cases were not being run, but I think it would be great to have them available in parallel with our current way to write tests. I didn't find a clean way to search all modules in g2p and run their doctests, so instead I added a test case in `test_utils.py` to run the doctests in g2p.mappings.utils, which does the trick here.

I like this hybrid model, but I don't know if you do, so I decided to make this a PR on my PR so you could review it separately.

There are also a few irrelevant trivial changes recommended by pylint, I assume they'll be uncontroversial.


g2p/tests/test_utils.py
```diff
@@ -30,11 +31,20 @@ def tearDown(self):
...
+        with open(
+            os.path.join(PUBLIC_DIR, "mappings", "generated_add.yaml"),
+            "w",
+            encoding="utf8",
+        ) as f:
             yaml.dump(
                 fresh_config, f, Dumper=utils.IndentDumper, default_flow_style=False
             )
 
+    def test_run_doctest(self):
```
roedoejet, line 43: Nice! yes, I like this. I'm generally a fan of doctests so it would be nice to include them in the tests here and start adding them.

Is there really no way to just find and run all doctests on the entire `g2p` library?

roedoejet (approved):

g2p/tests/test_utils.py
joanise: There must be, there just has to be, but from https://www.google.com/search?q=discover+doctest+suite+in+python I got to https://stackoverflow.com/questions/16982514/python-test-discovery-with-doctests-coverage-and-parallelism where they're asking the right question but the answer is more-or-less useless.

joanise: `doctest.DocTestSuite()` seems like the obvious answer, but I didn't find a way to make it discover everything under a given module and its children. I've only been able to make it find the tests by explicitly giving it the module. I'm not keep on writing this:
```
import g2p
import g2p.mappings
import g2p.mappings.utils
...
tests.addTests(doctest.DocTestSuite(g2p))
tests.addTests(doctest.DocTestSuite(g2p.mappings))
tests.addTests(doctest.DocTestSuite(g2p.mappings.utlis))
...
```
but I didn't find the correct way to not have to write this all out explicitly yet.

joanise: But yeah, I love the doctest syntax, and I'd use them by default for simple utility functions if we could automate their discovery.

PR #303, opened 2023-11-15
kwk updates discussed with cedar 

joanise:
Agreed with Daisy Rosenblum and her lab CEDaR:
-	I will automate the chaining of the equiv mapping for Umista, i.e., converting from kwk-umista to kwk-ipa will now first apply umista_equiv.csv, then kwk_umista_to_ipa.json. This also means that ReadAlong Studio will automatically recognize and process a broader set of inputs, in particular all the different diacritic variants covered by the equiv mapping.
-	We will make a change to the umista equiv mapping, and the umista->IPA mapping, to use U+0315 (combining comma above right) instead of U+0313 (combining comma above) as the standard diacritic in the digraphs k̕w, t̕ɬ, t̕s, and ḵ̕w since you said that’s what the community prefers to use. (But with the chaining of equiv, either will be accepted when converting from kwk-umista to kwk-ipa.)
-	All the kwk related mappings will be made case-insensitive, i.e., a letter means the same thing whether it’s upper case or lower case (but small caps remain different).
-	BOAS->Umista as well as the equiv mappings will be case preserving, so the output follows the same casing pattern as the input.
-	Since your lab discovered and uses the hidden feature that “g2p convert filename.txt inlang outlang” reads the contents of filename.txt instead of treating that argument as the text itself, I will unhide this feature and mention it in the documentation. E.g., one can do `g2p convert filename.txt kwk-umista kwk-umista-equiv > filename.equiv.txt`.
-	I’m not sure why the language code for umista equiv is “kwk-umista-con” instead of “kwk-umista-equiv”. I propose to rename it to that, so it’s more consistent with how other languages are configured. The equiv mapping for every other language in the g2p library is called langcode-equiv, and I think kwk should follow that pattern. -- Proposal accepted


joanise:
@roedoejet This PR should be tested and reviewed by CEDaR first. I will e-mail them.

joanise:
Closing this PR since I've updated it on another branch. Once Daisy reviews the new branch or a further update of it, I'll submit a new PR and merge then.

PR #305, opened 2024-01-04
perf: only create APP when it is really needed

joanise:
This speeds up g2p -h by a couple seconds on my machine.

It continues to bother me that `g2p -h` takes about 7 seconds, now 5 seconds with this PR. It should be sub-second, but it's taking refactoring I don't think we want to do it. However, this small patch is a first step.

### DIFF
--- a/g2p/cli.py
+++ b/g2p/cli.py
@@ -17,7 +17,6 @@ from networkx import has_path
 from g2p import make_g2p, make_tokenizer
 from g2p._version import VERSION
 from g2p.api import update_docs
-from g2p.app import APP
 from g2p.exceptions import InvalidLanguageCode, MappingMissing, NoPath
 from g2p.log import LOGGER
 from g2p.mappings import MAPPINGS_AVAILABLE, Mapping, MappingConfig
@@ -57,6 +56,8 @@ if "pytest" not in sys.modules:  # pragma: no cover
 
 def create_app():
     """Return the flask app for g2p"""
+    from g2p.app import APP
+
     return APP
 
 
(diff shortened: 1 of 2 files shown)


joanise:
Coverage goes down because 100% of the one added lines is not exercised: indeed, when we test the `APP` in the test suite, we do it via gunicorn rather than calling this `create_app()` function where I tucked the import.

dhdaines:
There should ... in theory ... be a way to test that line, but it doesn't seem to be really important.  I also agree that it should be instantaneous, it feels like g2p is already written in JavaScript, and yet we can't run it in the browser :)

joanise:
I do have a question: do we ever use (and do we still need) `g2p routes`, `g2p run` and `g2p shell`? Instead of `g2p run` we normally use `run_studio.py` or `gunicorn`, and `g2p routes` is half replaced by http://localhost:5000/docs (although I used `g2p routes` to figure that out...)

If we remove the Flask app altogether from the CLI and keep only the regular cli operation in the CLI, then I can get it to load nearly instantly relatively easily.

roedoejet:
> I do have a question: do we ever use (and do we still need) `g2p routes`, `g2p run` and `g2p shell`? Instead of `g2p run` we normally use `run_studio.py` or `gunicorn`, and `g2p routes` is half replaced by http://localhost:5000/docs (although I used `g2p routes` to figure that out...)
> 
> If we remove the Flask app altogether from the CLI and keep only the regular cli operation in the CLI, then I can get it to load nearly instantly relatively easily.

I personally don't have any use for `routes`, `run` or `shell` so this seems reasonable if it results in a performance bump like you say.

joanise:
> There should ... in theory ... be a way to test that line, but it doesn't seem to be really important

It turned out it wasn't hard, so I just pushed a second commit doing it.

PR #306, opened 2024-01-04
Remove the Flask app from the CLI to speed it up

joanise:
This is a pretty significant refactoring effort to move all expensive imports out of two places:
 - `g2p/__init__.py`: just importing `g2p` should not trigger loading and importing everything
 - `g2p/cli.py`: just calling `g2p -h` or `g2p cmd -h` should not trigger loading and importing everything

In both cases, the imports are tucked away inside the functions that need them.

However, we have constants and interfaces we need higher up, so I moved those to a new `g2p/constants.py` file. It's not super elegant, but I think it's at least a partial design improvement: instead of making a top-level module need a deeper one just for its constants or its class definition for typing, I moved them into a lightweight file that both import. This is definitely up for discussion and potential improvements.

But... `g2p -h` now takes just 0.9 seconds on my machine, and a mere 0.05 seconds in the CI run that just finished a few minutes ago.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,6 +46,28 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: false # too many upload errors to keep "true"
+      - name: Make sure the CLI stays fast
+        id: cli-load-time
+        run: |
+          PYTHONPROFILEIMPORTTIME=1 g2p -h 2> importtime.txt > /dev/null
+          CLI_LOAD_TIME="$((/usr/bin/time --format=%E g2p -h > /dev/null) 2>&1)"
+          echo "CLI load time: $CLI_LOAD_TIME"
+          cat importtime.txt
+          echo "CLI load time: $CLI_LOAD_TIME" > import-message.txt
+          echo "PR head ${{ github.event.pull_request.head.sha }}" >> import-message.txt
+          echo "Imports that take more than 0.1 s:" >> import-message.txt
+          grep -E 'cumulative|[0-9]{6}' importtime.txt >> import-message.txt
+          if [[ "$CLI_LOAD_TIME" > "0:01.00" ]]; then \
+            echo "ERROR: g2p --help is too slow."; \
+            echo "Please run 'PYTHONPROFILEIMPORTTIME=1 g2p -h 2> importtime.txt; tuna importtime.txt' and tuck away expensive imports so that the CLI doesn't load them until it uses them."; \
+            false; \
+          fi
+      - name: Report help speed in PR
+        if: github.event_name == 'pull_request'
+        uses: mshick/add-pr-comment@v2
+        with:
+          preformatted: true
+          message-path: import-message.txt
 
   test-on-windows:
     # Make sure stuff stays compatible with Windows by testing there too.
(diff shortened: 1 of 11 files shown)


.github/workflows/tests.yml
```diff
@@ -46,6 +46,28 @@ jobs:
...
+        run: |
+          PYTHONPROFILEIMPORTTIME=1 g2p -h 2> importtime.txt > /dev/null
+          CLI_LOAD_TIME="$((/usr/bin/time --format=%E g2p -h > /dev/null) 2>&1)"
+          echo "CLI load time: $CLI_LOAD_TIME"
+          cat importtime.txt
+          echo "CLI load time: $CLI_LOAD_TIME" > import-message.txt
+          echo "PR head ${{ github.event.pull_request.head.sha }}" >> import-message.txt
+          echo "Imports that take more than 0.1 s:" >> import-message.txt
+          grep -E 'cumulative|[0-9]{6}' importtime.txt >> import-message.txt
+          if [[ "$CLI_LOAD_TIME" > "0:01.00" ]]; then \
```
dhdaines, line 60: wow, does this work?

g2p/constants.py
```diff
@@ -0,0 +1,43 @@
...
+LANGS_FILE_NAME = "langs.json.gz"
+NETWORK_FILE_NAME = "network.json.gz"
+
+
+class BaseTransducer(ABC):
+    """Base class to typecheck transducers without having to import them."""
+
+    @abstractmethod
+    def __call__(self, to_convert: str):
+        pass
```
dhdaines: You could add a `#pragma: no cover` here,  I suppose, or just a bogus test that calls this abstract method and ensures that it does nothing and returns `None`.

Likewise below (I'm not going to add another comment just to say the same thing)

g2p/constants.py
```diff
@@ -0,0 +1,43 @@
...
+    "fast_levenshtein_distance",
+    "levenshtein_distance",
+]
+
+LANGS_DIR = os.path.join(os.path.dirname(__file__), "mappings", "langs")
+LANGS_FILE_NAME = "langs.json.gz"
+NETWORK_FILE_NAME = "network.json.gz"
+
+
+class BaseTransducer(ABC):
```
dhdaines, line 21: I mean strictly speaking these are not *constants*, so you could decide if it's worth the trouble to put them in another separate module called `g2p.types` or `g2p.abc`.

dhdaines (approved):
Looks good to me, comments are optional changes.

.github/workflows/tests.yml
joanise: Yeah, it's sweet. We've been using this in EveryVoice, that's where I copied it from.

g2p/constants.py
joanise: That's a good point. I didn't like the abuse of the name, `g2p.types` would be a lot clearer.

.github/workflows/tests.yml
joanise: OK, it's a bit clunky, but in CI I don't mind that too much, and the message you get in the PR conversation is really nice.

As for `"$CLI_LOAD_TIME" > "0:01.00"` doing the right thing, I force the time output to that syntax with `/usr/bin/time --format=%E` higher up.

g2p/constants.py
joanise: Refactored now.

roedoejet (approved):
Thanks for doing this @joanise ! I tested this locally and `g2p -h` took 0.076s 🚀

PR #308, opened 2024-01-19
fix(deps): lock numpy<2 because 2.0.0 is coming and has breaking changes

joanise:
See: https://pythonspeed.com/articles/numpy-2

When numpy 2.0.0 is actually out, we'll want to test g2p and our other projects against it before allowing it.


roedoejet (approved):
lgtm!

PR #313, opened 2024-01-31
Move southern and Northern after East Cree to improve sorting

joanise:
Per community request some months ago.

Result in Studio-Web:
image


joanise:
@deltork This is something you asked for back in September...

PR #315, opened 2024-02-01
migrate deprecated pydantic v1 style validators to pydantic v2 recommended alternative

joanise:
also:
 - configure mypy for pydantic v2
 - make the test runners DRYer


joanise:
@roedoejet Your unit tests were nice for this: they caught the problems while my solution was still not completely functional.

joanise:
@roedoejet Changes made, ready to rereview.

g2p/mappings/utils.py
```diff
@@ -803,30 +803,34 @@ def validate_preserve_case(self):
...
-            display_name = f"{values['in_lang']} {find_mapping_type(values['in_lang'])} to {values['out_lang']} {find_mapping_type(values['out_lang'])}"
-        return display_name
+    @field_validator(
+        "rules_path", "abbreviations_path", "alignments_path", mode="before"
+    )
+    @classmethod
+    def add_parent_dir(cls, value: Any, info: ValidationInfo):
+        """If there is a parent directory, prepend it to all path fields."""
+        parent_dir = info.data.get("parent_dir", None)
+        if value is not None and parent_dir:
```
roedoejet, line 813: since value is type Any here, it might be good to check `isinstance(value, string)` or something, maybe a try/except here? If someone passes an invalid type to any of the fields validated by this validator it will fail without a proper `pydantic.ValidationError` since the mode is `before`

roedoejet (changes requested):
minor change: just ensure we handle errors to`Path(parent_dir) / value` in `mode='before'` field validators.

g2p/mappings/utils.py
joanise: That's quite disappointing. The field is still typed  `Optional[Path]`, so I expect that validation to still happen at some point during the processing,

But OK, fixing it.

joanise:
OK, that was harder than it should have been. I think it's good now, though. Ready, once again, for re-review.

g2p/mappings/__init__.py
```diff
@@ -432,8 +432,11 @@ def load_mapping_config_from_path(path_to_mapping_config: Union[str, Path]):
                     mapping["parent_dir"] = parent_dir
         try:
             return MappingConfig(**loaded_config)
+        except exceptions.MalformedMapping as e:
+            e.message = f"{e.message}\nProblem in config file: {path}"
+            raise e
```
joanise, line 437: This code is currently unreachable (we never raise MalformedMapping now that we're using Pydantic) but I'm going to leave it here because I want to figure out, for #320, how to catch the pydantic exception and raise a MalformedMapping from it to full it up here with a message telling the user which file the problem occurs in, with the pydantic error saying which field is the problem.

roedoejet: agreed - we can improve on these. thanks for raising the issue @joanise !

roedoejet (approved):
This is a really good, thorough re-write of those validators thanks @joanise !

PR #316, opened 2024-02-06
Move to the current stable black release. 23.x

joanise:
The code is already compliant with that format, but folks will have to redo `pip install -r requirements/requirements.dev.txt` so future code continues to be compliant.


roedoejet (approved):
sounds good

Issue #322, opened 2024-02-13
Creating an empty mapping corrupts the langs.json.gz file


joanise:
Steps to reproduce:
 - create a new mapping with an empty file (i.e., a set of zero rules), e.g., `g2p/mappings/langs/lang/lang_equiv.csv` and configure it.
   (alternatively: `cat /dev/null > g2p/mappings/langs/oka/oka_equiv.csv`)
 - run `g2p update`
 - run any CLI command that causes langs to be loaded, e.g. `g2p convert a fra fra-ipa` and get a stack trace ending with 
   > FileNotFoundError: [Errno 2] No such file or directory: 'lang_to_lang-equiv.json'

Use case for these steps: you intend to populate it later but it's a placeholder for lang-equiv.csv, which you added right away to `config-g2p.yaml` even though you didn't write the equiv rules yet.

Populating the empty csv file with rules (or restoring  `oka_equiv.csv`) and re-running `g2p update` won't fix it, since `g2p update` fails with the same error. Only `git restore g2p/mappings/langs/langs.json.gz` will allow recovery at this point.

Issue #320, opened 2024-02-09
Pydantic error messages for problems in config-g2p files are unfriendly


joanise:
Before the conversion to Pydantic, problems in the config files would raise `MalformedMapping` with an (at least somewhat) informative message.

Now we get a pydantic error that's much harder to interpret. We should try to catch it it the right place and augment it with user-informative details.

E.g., set `abbreviations_path: 1.0` in any mapping config and running `g2p udpate` yields this:
```
$ g2p update
INFO - Loading alignments from c:\users\joanise\sandboxes\g2p\g2p\mappings\langs\eng\cmudict.ipa.aligned.txt
Traceback (most recent call last):
  File "C:\Users\joanise\Miniconda3\envs\ras-py310\Scripts\g2p-script.py", line 33, in <module>
    sys.exit(load_entry_point('g2p', 'console_scripts', 'g2p')())
  File "C:\Users\joanise\Miniconda3\envs\ras-py310\lib\site-packages\click\core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "C:\Users\joanise\Miniconda3\envs\ras-py310\lib\site-packages\click\core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "C:\Users\joanise\Miniconda3\envs\ras-py310\lib\site-packages\click\core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Users\joanise\Miniconda3\envs\ras-py310\lib\site-packages\click\core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Users\joanise\Miniconda3\envs\ras-py310\lib\site-packages\click\core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "c:\users\joanise\sandboxes\g2p\g2p\cli.py", line 662, in update
    cache_langs(dir_path=in_dir, langs_path=langs_path, network_path=network_path)
  File "c:\users\joanise\sandboxes\g2p\g2p\mappings\langs\utils.py", line 163, in cache_langs
    mapping_config = MappingConfig.load_mapping_config_from_path(path)
  File "c:\users\joanise\sandboxes\g2p\g2p\mappings\__init__.py", line 434, in load_mapping_config_from_path
    return MappingConfig(**loaded_config)
  File "C:\Users\joanise\Miniconda3\envs\ras-py310\lib\site-packages\pydantic\main.py", line 171, in __init__
[...]

PR #323, opened 2024-02-13
Detect empty mappings and fail early.

joanise:
Empty mappings cause the g2p database to get corrupted, so block them.
Other changes were things I noticed while fixing this, including some coverage improvements.

Fixes: #322


g2p/mappings/__init__.py
```diff
@@ -347,7 +347,10 @@ def export_to_dict(self, mapping_type="json", config_only=False):
                 exclude_none=True,
                 exclude={"parent_dir": True},
             )
-        model_dict["rules_path"] = f"{self.in_lang}_to_{self.out_lang}.{mapping_type}"
+        if not model_dict.get("rules_path"):
```
dhdaines, line 350: should be `if model_dict.get("rules_path") is None` in the off chance that it's a file named `False`? ;-)

g2p/mappings/utils.py
```diff
@@ -389,6 +389,10 @@ def load_from_file(path: Union[Path, str]) -> list:
         raise exceptions.IncorrectFileType(
             f"File {path} is not a valid mapping filetype."
         )
+    if not mapping:
+        raise exceptions.MalformedMapping(
```
dhdaines, line 393: also need a unit test to verify that this gets raised properly

dhdaines (changes requested):
just a couple of little  changes, otherwise good!

g2p/mappings/__init__.py
joanise: You've got to be kidding! These file formats can be a nuisance sometimes, but you're right, I've triggered bugs by having `1.0` or `True` as a path elsewhere...
But I also want to fail on "", should the user have typed `rules_path: ` without a value.

So I guess to be safe I'd need to test explicitly for None or "".

Ah-ha, no! By the time we get here, the typing of `rules_path: Optional[Path]` has already been validated by Pydantic and `True` will have triggered a validation error earlier. And an ugly one at that, see #320, which I want to fix.

g2p/mappings/utils.py
joanise: Good point. This is the very bug that triggered this line of work, but I didn't make a unit test for it. Thanks for pointing it out!

joanise:
@dhdaines Unit test added, ready to re-review.

dhdaines (approved):

PR #330, opened 2024-03-04
docs: minor improvements to g2p manual

joanise:
g2p.mappings.Rule documentation:
 - show the fields in the logical order with "members_order: source"
 - show the fields in their normal case rather than all caps with "heading_level: 3" instead of 4

correct a minor detail in the background description of the project.


roedoejet (approved):
looks good!

PR #334, opened 2024-03-05
Fully revamp the release workflow, with PEP440 versions and publication triggered by a tag push

joanise:
This is currently "disabled" with the real actions being commented out or prefixed with "echo".

I'd like to test this by pushing an actual test tag, but I want code review before I do it, to make sure I correctly disabled the real work while keeping enough for us to see the result.

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -1,36 +1,131 @@
-# This workflows will upload a Python Package using Twine when a release is created
-# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+# To create a release:
+#  - update the version number in g2p/_version.py
+#  - commit the changes
+#  - create an annotated tag with the version number, e.g.: git tag -a v2.0.1 -m "v2.0.1"
+#  - push the tag, which will trigger this pythonpublish release workflow
 
-name: Upload Python Package
+name: Publish g2p to PyPI and create a GitHub release
+
 on:
   push:
-    branches: [ release ]
+    tags:
+      - v[0-9]+.**
+
 jobs:
-  deploy:
+  tests:
+    uses: ./.github/workflows/tests.yml
+    secrets: inherit
+
+  build-docs:
+    uses: ./.github/workflows/docs.yml
+    needs: tests
+    secrets: inherit
+
+  build:
     runs-on: ubuntu-latest
+    needs: tests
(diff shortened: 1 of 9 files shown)


.github/workflows/pythonpublish.yml
```diff
@@ -1,36 +1,139 @@
...
+      url: https://pypi.org/p/g2p
+    permissions:
+      id-token: write  # IMPORTANT: mandatory for trusted publishing
+    steps:
+    - name: Download the distribution packages
+      uses: actions/download-artifact@v4
+      with:
+        name: python-package-distributions
+        path: dist/
+    - name: Publish distribution 📦 to TestPyPI
```
roedoejet, line 120: secrets.TEST_PYPI_API_TOKEN isn't available on g2p. I'll create one for secrets.PYPI_API_TOKEN but I don't think I have a test pypi account for making a PAT

roedoejet (approved):
This is awesome @joanise - nicely done. I'll add the PAT for Pypi, but we'll also need to add a test pypi project and PAT if we want the test pypi job to run. I don't think it's necessary really though. we already know g2p works on pypi

.github/workflows/pythonpublish.yml
dhdaines: I have a test pypi account if you need someone to test this...

dhdaines: In fact, I can just add g2p right now as a trusted publisher!

dhdaines: If you want to change it to use trusted publishing, you can publish to the test.pypi.org project "gi2pi" (the name "g2p" was already taken).

Otherwise, I have an API token but I don't have permission to add it as a secret to this repository...

roedoejet: oh nice - I didn't know you could do that. thanks @dhdaines !

joanise: I agree, Aidan, we can drop test PyPI.
dhd, that would be great if we can use trusted publishing, thanks!

joanise:
This workflow is now tested and fixed, and show work. There were a few bugs I discovered by pushing a fake v2.0.0a0 tag, which I have now deleted. I had the workflow set not to do (non-reversible) real work, just to be safe. But now it'll do the real thing when we push v2.0.0.
Please re-review @roedoejet @dhdaines to be on the safe side, but I'm fairly confident it's all good now.

PR #339, opened 2024-03-19
Add a basic migration guide to v2

### LINKED ISSUE #325: Write up a migration guide for converting to g2p 2.0
2.0 has a number of breaking changes from 1.x, so we need to provide a migration guide.

dhdaines:
Migration guide to v2.  Makes it clear enough (I hope) that internal APIs may still be subject to change!

Fixes #325

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,10 +204,11 @@ You can also run the `g2p Studio` which is a web interface for creating custom l
 ## API for Developers
 
 There is also a REST API available for use in your own applications.
-To launch it from the command-line use `flask --app g2p.app run`.  The
-API documentation will be viewable (with the ability to use it
-interactively) at http://localhost:5000/docs - an OpenAPI definition
-is also available at http://localhost:5000/static/swagger.json .
+To launch it from the command-line use `python run_studio.py` or
+`flask --app g2p.app run`.  The API documentation will be viewable
+(with the ability to use it interactively) at
+http://localhost:5000/docs - an OpenAPI definition is also available
+at http://localhost:5000/static/swagger.json .
 
 You can see the list of URLs served by the API using `flask --app
 g2p.app routes`.
--- a/README.md
+++ b/README.md
@@ -201,6 +201,17 @@ Once you have the Compiled DB, it is then possible to use the `g2p convert` comm
 
 You can also run the `g2p Studio` which is a web interface for creating custom lookup tables to be used with g2p. To run the `g2p Studio` either visit https://g2p-studio.herokuapp.com/ or run it locally using `python run_studio.py`.
 
+## API for Developers
+
+There is also a REST API available for use in your own applications.
+To launch it from the command-line use `flask --app g2p.app run`.  The
+API documentation will be viewable (with the ability to use it
+interactively) at http://localhost:5000/docs - an OpenAPI definition
+is also available at http://localhost:5000/static/swagger.json .
+
+You can see the list of URLs served by the API using `flask --app
+g2p.app routes`.
+
 ## Maintainers
 
 [@roedoejet](https://github.com/roedoejet).
--- a/README.md
+++ b/README.md
@@ -201,8 +201,6 @@ Once you have the Compiled DB, it is then possible to use the `g2p convert` comm
 
 You can also run the `g2p Studio` which is a web interface for creating custom lookup tables to be used with g2p. To run the `g2p Studio` either visit https://g2p-studio.herokuapp.com/ or run it locally using `python run_studio.py`.
 
-Alternatively, you can run the app from the command line: `g2p run`
-
 ## Maintainers
 
 [@roedoejet](https://github.com/roedoejet).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -201,7 +201,16 @@ Once you have the Compiled DB, it is then possible to use the `g2p convert` comm
...
 You can also run the `g2p Studio` which is a web interface for creating custom lookup tables to be used with g2p. To run the `g2p Studio` either visit https://g2p-studio.herokuapp.com/ or run it locally using `python run_studio.py`.
 
-Alternatively, you can run the app from the command line: `g2p run`
+## API for Developers
+
+There is also a REST API available for use in your own applications.
+To launch it from the command-line use `flask --app g2p.app run`.  The
+API documentation will be viewable (with the ability to use it
+interactively) at http://localhost:5000/docs - an OpenAPI definition
+is also available at http://localhost:5000/static/swagger.json .
```
joanise, line 210: These two links also work when we run `python run_studio.py`. Is there an advantage to running it via `flask --app` instead of via our our script? In any case, to avoid confusion since the two commands are effectively equivalent, I would say here 

    To launch it from the command-line use `flask --app g2p.app run` or `python run_studio.py`.

so that it's clear you get the same thing either way.

Both commands actually spin up the g2p-studio on http://localhost:5000 so it'll be confusing otherwise.

joanise: Oh, the API docs and swagger are also available on https://g2p-studio.herokuapp.com/ but I don't know if we need to mention that here.

docs/migration-2.md
```diff
@@ -0,0 +1,119 @@
+---
+comments: true
+---
+
+# Migrating from `g2p` 1.0
```
joanise, line 5: 1.x?

docs/migration-2.md
```diff
@@ -0,0 +1,119 @@
...
+not change anything, but if you relied on internal classes and
+functions from `g2p.mappings`, `g2p.transducer`, etc, then you can no
+longer depend on them being also accessible in the top-level `g2p`
+package.  For example, you will need to make this sort of change:
+
+```diff
+- from g2p import Mapping, Transducer, make_tokenizer
++ from g2p.mappings import Mapping
++ from g2p.transducer import Transducer
++ from g2p.mappings.tokenizer import make_tokenizer
```
joanise, line 56: `make_tokenizer` is also part of the official documented API, and it should be imported as `from g2p import make_tokenizer`. You'll notice in `g2p/__init__.py` that it's just a dummy wrapper, but it has the major advantage of deferring the import to actual use, which means this is the recommended way to import that function for packages using g2p as an external module.

This section is correct for other things, like `Mapping` and `Transducer`.

mkdocs.yml
```diff
@@ -35,6 +35,7 @@ nav:
     - Getting started: start.md
     - How to contribute: contributing.md
     - Using the g2p studio: studio.md
+    - Migrating from g2p 1.0: migration-2.md
```
joanise, line 38: 1.x?

Or maybe "Migrating from g2p 1.0 or 1.1"?

docs/migration-2.md
joanise: Or "1.0 or 1.1"?

joanise (changes requested):
This is great, thanks!
I have a number of small changes to request, but it's nice and clear overall, and catches things I did not realize. And taught me the use of the `flask --app` CLI command.

README.md
dhdaines: I figured I'd leave that out since if people are develop-y enough to use the API then maybe they can self-host...

docs/migration-2.md
```diff
@@ -0,0 +1,120 @@
...
+which, unfortunately, are incompatible with mappings and Python code
+written for the previous version.  We have tried to describe them here
+with the changes you will need to make to your code and data.
+
+## Mapping configurations have changed (for the better)
+
+The configurations for mappings (which you'll find in
+`g2p/mappings/langs/*/config-g2p.yaml`) are now validated with a [YAML
+Schema](https://raw.githubusercontent.com/roedoejet/g2p/main/g2p/mappings/.schema/g2p-config-schema-2.0.json).
+If you use an editor like [Visual Studio
```
roedoejet, line 17: I think the [yaml extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml) needs to be installed to have it work on VS code

docs/migration-2.md
```diff
@@ -0,0 +1,120 @@
...
+
+```python
+mapping = Mapping.load_from_file("path/to/some/config-g2p.yaml")
+print("Case sensitive?", mapping.case_sensitive)
+```
+
+To iterate over the rules in a mapping, you now use the `rules`
+property instead of the `mapping_data` field.  The rules themselves
+now also use properties for access, which do not entirely correspond
+to the names used in the JSON definition, because `in`, for example,
```
roedoejet, line 84:
```suggestion
to the names used in the JSON/YAML rule definition files, because `in`, for example,
```

roedoejet (approved):
This is really awesome @dhdaines ! Thank you!!! I've made a few small comments/suggestions, but everything looks great to me otherwise.

joanise (approved):
Looks great, thanks!

PR #343, opened 2024-03-20
style: apply consistent 2 indent in all ci workflows

joanise:
Our indentation was inconsistent from one workflow file to another.

I like this guide, 
https://developers.home-assistant.io/docs/documenting/yaml-style-guide/#sequences
which we were already mostly following except for having the list of steps not indented in half the workflows.


roedoejet (approved):
lgtm

PR #348, opened 2024-03-25
Fix Atikamekw mapping of tc

dhdaines:
Seems that it has always been broken ... oops!


joanise (approved):
Looks good to me, thanks for adding relevant test data.

Issue #351, opened 2024-04-02
g2p convert added blank line at end. 


marctessier:
When I run  `g2p convert ` my output file always has a blank line at end.

Supper annoying, I always have to remember to run `sed -i '$ d' *.gp2` to remove afterwards.

joanise:
@marctessier Let me know if #352 fixes this as you'd like.

marctessier:
Awesome, yes that resolved the issue  . I am not seeing the extra empty  line at end!

PR #352, opened 2024-04-02
fix: g2p convert should not add newline when input is a file

joanise:
Fixes: #351


roedoejet (approved):
lgtm! thanks @joanise !

PR #355, opened 2024-04-02
Add mapping for Makah

dhdaines:

- [x] `config-g2p.yaml` file includes all author names, and settings necessary
- [x] You agree to license your contribution under the same license as this project (see [LICENSE](https://github.com/roedoejet/g2p/blob/main/LICENSE) file).

* **Other information**:

Based on the official desktop keyboard layout (https://makahmuseum.com/makah-keyboard/) from the Makah Language Program (https://makahmuseum.com/departments/makah-language-program/)

Note that we also accept the orthgraphy from [Davidson 2002](https://depts.washington.edu/wll2/files/davidson_02_diss.pdf) (which is really Nuuchahnulth) - differences from the standard orthography are:

- the voiceless uvular fricative is x̌ in Makah orthography versus x̣ in Davidson
- vowel length is indicated by · (center dot) and : (colon) versus U+02D1 ˑ MODIFIER LETTER HALF TRIANGULAR COLON and  U+02D0 ː MODIFIER LETTER TRIANGULAR COLON in Davidson
- the voiceless lateral fricative is ɫ (U+026B LATIN SMALL LETTER L WITH MIDDLE TILDE) versus ɬ (U+026C LATIN SMALL LETTER L WITH BELT) in Davidson (and, also [Jacobsen 1999](https://books.google.ca/books/about/First_lessons_in_Makah.html?id=MB7fSgAACAAJ&redir_esc=y))


g2p/mappings/langs/font-encodings/myh_doulos.json
```diff
@@ -0,0 +1,282 @@
+[
+  {
+    "in": "\\ ",
```
roedoejet, line 3: What are all these \\ rules about?

g2p/mappings/langs/font-encodings/myh_doulos.json
```diff
@@ -0,0 +1,282 @@
...
+  {
+    "in": "t",
+    "out": "u"
+  },
+  {
+    "in": "u",
+    "out": "P"
+  },
+  {
+    "in": "v",
```
roedoejet, line 267: Why are there ASCII characters mapped to other ASCII characters as part of this mapping? Surely that's not part of the font-encoding?

roedoejet: I see prevent-feeding is true globally. Is it really the case that this font renders `s` as `t` and `t` as `u`?? Why!?

roedoejet (commented):
Thanks @dhdaines - I've asked a couple clarifying questions about the rules with backslashes and ASCII-to-ASCII rules.

g2p/mappings/langs/font-encodings/myh_doulos.json
dhdaines: Ah, that is an artifact of the rules being semi-automatically generated from the character map in the PDF.  I'll remove it!

g2p/mappings/langs/font-encodings/myh_doulos.json
dhdaines: Amazingly enough, yes - try copying some text from https://depts.washington.edu/wll2/files/davidson_02_diss.pdf - no idea if something went weird in that document in particular, if you have some other Nuuchahnulth documents with SIL fonts in them it would be good to check.

g2p/mappings/langs/font-encodings/myh_doulos.json
dhdaines: but also, we need backslashes because the 'in' is a regex

dhdaines: (I really don't *like* that it's a regex because it means we can't generally express g2p rules as FSTs, but it is what it is - perhaps we should define a subset of Python regex language that we accept)

g2p/mappings/langs/font-encodings/myh_doulos.json
dhdaines: The PDF mapping to "Unicode" characters is very weird, no idea at all how it got that way:
Screenshot from 2024-04-02 07-46-50

joanise (approved):
I can't judge the mapping itself, except for the fact that it looks a lot like some of our other APA-based mappings, but this PR looks good to me in every way I can fairly assess it.

PR #357, opened 2024-04-03
Improve coverage and type checking slightly

dhdaines:
Some things from `dev.hatch` that do not depend on `hatch`


joanise (approved):
I can't really tell what those type: ignore pragmas change, but sure.
The additional testing is certainly useful.

dhdaines:
They are for modules that `mypy` can't type-check, which causes it to fail with some errors (since pre-commit only runs `mypy` on changed files, we won't notice the problem unless we change files)

Issue #350, opened 2024-04-02
running g2p convert on  "large" files takes forever...


marctessier:
I had a 4k line files I wanted to `g2p convert ` . After waiting for a very longs time  , I decided to split my input file into 100line chunks and try that while waiting on my original job to finish.    Processing the 40 x 100line chunks was completed  in no time compared to processing the 4k original that was still processing.   

Something is wrong or working very inefficiently.

joanise:
Oh boy, you're extending the intent of this functionality in a serious way!
The fact that we accept a file at all, instead of only text on the command line, was a hidden hack, which only became documented because some users discovered it by accident.
Now you want it to be efficient! LOL

The issue is probably due to the fact we load the entire file in memory and then process it. If large files is a real use case, we could convert it to line-oriented processing.

dhdaines:
I think it's more likely due to the way we apply rules which is really quite "pessimal" as each rule gets applied successively on the entire input, so O(MN) and if either M or N gets big then that can take a while.

If we didn't want to accept arbitrary Python regular expressions as inputs, we could compile them into a minimal deterministic FST ;-)

joanise:
@marctessier I get it now. I just did `g2p convert filelist-text-only.txt eng eng-arpabet` for testing Aidan's big EV PR, and boom, it's taking a long time!

@dhdaines I would be very surprised if g2p was convertible to an FST. I imagine your big-Oh analysis is in the right ballpark, and clearly line-oriented (or token-oriented, as you suggest) processing will make a big difference. I guess token would be best, since that's less likely to trip on long lines, but line is easiest to implement...

joanise:
@dhdaines I think the way we build the transduction graph, by inserting and removing nodes and edges, is going to be part of the issue: each operation probably costs O(N), with N ~ the length of the input.

joanise:
@marctessier You'll appreciate #358

PR #358, opened 2024-04-03
perf: make g2p convert line-oriented when processing a file

joanise:
significant speed up when using it to process long files.

Fixes: #350

### DIFF
--- a/g2p/cli.py
+++ b/g2p/cli.py
@@ -531,10 +531,13 @@ def convert(  # noqa: C901
         raise click.UsageError(
             f"Path between '{in_lang}' and '{out_lang}' does not exist"
         )
+    # Figure if the input text is on the command line or in a file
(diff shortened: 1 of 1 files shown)


dhdaines (approved):
Looks like an easy win - you could even do `lines = open(input_text, encoding="utf8")` instead of reading all the lines at once, although then you have to close the file, so it's probably not worth it.

joanise:
> Looks like an easy win - you could even do `lines = open(input_text, encoding="utf8")` instead of reading all the lines at once, although then you have to close the file, so it's probably not worth it.

Yeah, I wanted my context manager to close the file in all exceptional cases, and I didn't want to have to wrap the whole rest of the function in a try/finally block, so I decided for this solution. We can review it if someone complains about memory at some point...

PR #359, opened 2024-04-04
API v2, second time around

dhdaines:
Here's the new API used in the alignment demo (https://github.com/dhdaines/alignment-demo and https://aligner.ecolingui.ca/), updated for G2P 2.x.  There's also a small fix to the substring alignment code, can't quite remember what it was for but it should have coverage in any case. (EDIT: Nope actually it doesn't have coverage ... well, that will have to get fixed)

For the moment I made fastapi an optional dependency, so this should have no effect on anything else.  It is based on `dev.hatch` though, at least for the moment.

### DIFF
--- a/g2p/api_v2.py
+++ b/g2p/api_v2.py
@@ -2,7 +2,7 @@
 
 You can run the API app for development purposes on any platform with:
     pip install uvicorn fastapi
-    DEVELOPMENT=1 uvicorn g2p.api_v2:app --reload --port 5000
+    DEVELOPMENT=1 uvicorn g2p.api_v2:api --reload --port 5000
 - The --reload switch will watch for changes under the directory where it's
   running and reload the code whenever it changes.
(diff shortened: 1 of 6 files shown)


joanise:
Initial clarification question: is the intent to ultimately merge both #360 and this PR, i.e., to update API v1 to use FastAPI, *and* to add this API v2?

dhdaines:
> Initial clarification question: is the intent to ultimately merge both #360 and this PR, i.e., to update API v1 to use FastAPI, _and_ to add this API v2?

Yes, they are independent of each other, but the idea would be to add both APIs (`/api/v1` and `/api/v2`)

joanise (approved):
Great!
Way easier to review this one, now that the hard work is done in the other PR!

dhdaines:
> Great! Way easier to review this one, now that the hard work was done in the other PR!

Thanks!  I'm not going to worry too much about the coverage issues for the moment...

PR #360, opened 2024-04-04
FastAPI port of API/Studio, second time around

dhdaines:
This time with backwards compatibility and testing! Wow!


dhdaines:
Note that fastapi-socketio is really broken, so it would probably be better to not use it: https://github.com/pyropy/fastapi-socketio/issues/52

In the end there's very little actual code in it so we can just use `socketio` directly.

dhdaines:
Also note that this would mean that the API and Studio require Python 3.8, so not sure if we want to go there...

joanise:
> Also note that this would mean that the API and Studio require Python 3.8, so not sure if we want to go there...

The Web API already requires Python 3.8, we only continue to support 3.7 for the programmatic Python API, i.e., g2p as a Python module.

dhdaines:
> > Also note that this would mean that the API and Studio require Python 3.8, so not sure if we want to go there...
> 
> The Web API already requires Python 3.8, we only continue to support 3.7 for the programmatic Python API, i.e., g2p as a Python module.

Ah, of course - then there's no problem since the API is a separate, optional dependency here.  But we need to separate the test suites so that we can test the non-API parts with 3.7.

dhdaines:
Also, not quite sure what's going wrong with coverage in CI here, it gives the correct results when I run it locally.

g2p/app.py
```diff
@@ -1,20 +1,33 @@
 """
+The g2p Studio web app.
 
-Views and config to the g2p Studio web app
+You can run the app (and API) for development purposes on any platform with:
+    pip install uvicorn
+    uvicorn g2p.app:APP --reload --port 5000
```
joanise, line 6: Trying that locally, and the locally spun up g2p-studio is all broken and generates tons of warnings at the console.
Was `pip install -e .[api]` and then this command the right thing to do?

g2p/app.py
```diff
@@ -123,30 +145,26 @@ def return_echart_data(tg: Union[CompositeTransductionGraph, TransductionGraph])
...
-@APP.route("/")
-def home():
+@APP.get("/", response_class=HTMLResponse)
+def home(request: Request):
     """Return homepage of g2p studio"""
-    return render_template("index.html")
+    return TEMPLATES.TemplateResponse("index.html", {"request": request})
 
 
-@APP.route("/docs")
```
joanise, line 132: I can never remember the whole `/api/v1/docs` path we need to get this info out of FastAPI (I usually remember the tokens, but have to try them in multiple orders before I get it right). Any chance we can `/docs` as a route that redirects to `/api/v1/docs`?

joanise: Besides being easier to remember, it would be more backwards compat.

joanise (changes requested):
I really like these changes, although on my end g2p-studio is broken on this branch, I don't know if I'm doing something wrong or if it's a bug.
I better get my act together and review your hatch PR, but having installed using hatch for this branch I'm already pretty much on-board now.

CI coverage: I don't know what's going on.

I've left notes about the clunky way I've been making sure we're still exercising Python 3.7 in CI. I'm open to better suggestions: the actual goal is that some part of CI should exercise g2p as a Python module and CLI command (i.e., all the tests that don't involve g2p-studio or the REST API) with Python 3.7. I'm not attached to how we get there, just that we do it. And that's a hint that my documentation was sorely inadequate, nothing in there pointed out that goal.

README.md
```diff
@@ -212,19 +211,19 @@ Once you have the Compiled DB, it is then possible to use the `g2p convert` comm
 
 ## Studio
 
-You can also run the `g2p Studio` which is a web interface for creating custom lookup tables to be used with g2p. To run the `g2p Studio` either visit https://g2p-studio.herokuapp.com/ or run it locally using `python run_studio.py`.
+You can also run the `g2p Studio` which is a web interface for
+creating custom lookup tables to be used with g2p. To run the `g2p
+Studio` either visit https://g2p-studio.herokuapp.com/ or run it
+locally with `run_studio.py`.
```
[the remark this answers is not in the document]
dhdaines: oh yes! good point!

docs/migration-2.md
```diff
@@ -100,22 +100,14 @@ been using them and should not be construed as public API documentation.
...
-
-    flask --app g2p.app routes
-    flask --app g2p.app shell
-
-To run G2P Studio, use:
+To run the `g2p` API server and G2P Studio, you can use:
 
     python run_studio.py
+
+It does not seem that any equivalents of `routes` or `shell` exist.
```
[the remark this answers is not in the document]
dhdaines: well, there's the auto-generated online API documentation, which is arguably more useful anyway...

g2p/app.py
dhdaines: ah oh hmm. will have to review the documentation to make sure it makes sense

g2p/app.py
dhdaines: Yes, this should be doable.  In fact arguably the top-level server should just use Starlette directly and not FastAPI (this would solve our fastapi-socketio issues) - then we can define whatever routes we like, including, I believe, pointing into the /docs for a mounted sub-application.  I'm curious to see if this works so I'll probably hack on it a bit this weekend.

dhdaines: (to clarify: the APIs are still FastAPI - but what it gives you is just an ASGI application, which can be "mounted" anywhere you like on Starlette)

joanise:
I believe this will be ready to merge, with #363 and a redirect link from `/docs` to `/api/v1/docs`, although when we also merge the v2, we might want a page listing both instead of an actual redirect...

g2p/app.py
joanise: Tested again, and I don't really see the problem I saw the first time. 
a) the warnings are on `origin/main` too, and are triggered by starting to type in the text box with the mapping set to Custom but no rules in there yet. I should make an issue of that.
b) this time I had done `pip install -e .[dev]` in this sandbox, and `uvicorn g2p.app:APP --reload --port 5000` seems to work just fine, so I'm not sure what happened when I first tested this.

run_studio.py
```diff
@@ -2,20 +2,22 @@
 
 import sys
 
-if sys.version_info < (3, 7, 0):  # pragma: no cover
+if sys.version_info < (3, 8, 0):  # pragma: no cover
```
joanise, line 5: Darn, I delete this block in my last PR, since it had become just 3.7 everywhere, so you'll get a conflict here and you'll have to bring it back. Sorry about that!

g2p/app.py
```diff
@@ -123,30 +145,26 @@ def return_echart_data(tg: Union[CompositeTransductionGraph, TransductionGraph])
...
-def docs():
-    """Return swagger docs of g2p studio API"""
-    return render_template("docs.html")
-
-
-@SOCKETIO.on("conversion event", namespace="/convert")
-def convert(message):
+@APP.sio.on("conversion event", namespace="/convert")  # type: ignore
+async def convert(sid, message):
```
joanise, line 155: That's weird, Codecov seems to think this function is not at all covered in CI testing. I find that hard to believe since it's the main function in the API! `test_studio.py` should trigger calls to this, right?

g2p/app.py
dhdaines: Okay, this is done now, and also we just use Starlette directly for the top-level app since FastAPI gives us nothing there (and in fact creates problems due to fastapi-socketio).  I use redirections to point /static/swagger.json and /docs to the same things as before.

g2p/app.py
dhdaines: yes something is weird with codecov, I'll need to fix that before merging

joanise: Actually, not sure it's codecov: I ran it locally with `coverage run` and got the same answer, which I still don't understand.

.github/workflows/tests.yml
```diff
@@ -92,7 +92,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.7"
+          python-version: "3.8"
```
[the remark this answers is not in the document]
dhdaines: should be fixed with your PR, I think!

g2p/app.py
dhdaines: It is probably an artifact of the switch to uvicorn and asyncio, which, apparently, doesn't really do signal handlers well.  Killing with SIGINT and waiting a bit longer *seems* to fix it.

.github/workflows/tests.yml
```diff
@@ -25,30 +25,26 @@ jobs:
...
       - name: Ensure browser is installed
         run: python -m playwright install --with-deps chromium
       - name: Launch the API
-        run: |
-          coverage run --parallel-mode \
-              -m gunicorn --worker-class eventlet -w 1 g2p.app:APP --no-sendfile --bind 127.0.0.1:5000 --daemon
+        run: coverage run run_studio.py &
       - name: Run dev suite
-        run: coverage run --parallel-mode run_tests.py dev
+        run: coverage run run_tests.py dev
```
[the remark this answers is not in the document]
dhdaines: I was surprised too!  But parallel mode comes from the configuration in `pyproject.toml` now so it's redudant here...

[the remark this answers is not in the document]
joanise: Oh, I see. That's cool!

joanise:
Meh, I'm being bitten by a stale environment again. I'll create something fresh to test.

joanise:
I hate dependency hell!

dhdaines:
> I hate dependency hell!

Hm, I'm really surprised this is happening ... I didn't have to change anything in my environment, the packages are the same as before the last updates except that we aren't using some of the ones we were previously...

joanise:
Found it:
```
     "starlette>=0.35.1",
-    "python-socketio>=5",
+    "python-socketio>=5.9.0",
     "uvicorn",
```

dhdaines:
> Found it:
> 
> ```
>      "starlette>=0.35.1",
> -    "python-socketio>=5",
> +    "python-socketio>=5.9.0",
>      "uvicorn",
> ```

aaah ... okay, I put python-socketio>=5 because that's what supports the same protocol as the JavaScript side, but I'll apply this fix (presuming there is no problem with Python 3.8)

joanise:
That's the minimum required patch. On a stale environment with `python-socketio==5.8.0` I got exceptions like this from `./run_studio.py`:

```
WARNING:  No supported WebSocket library detected. Please use 'pip install uvicorn[standard]', or install 'websockets' or 'wsproto' manually.
INFO:     127.0.0.1:62614 - "GET /ws/socket.io/?EIO=4&transport=websocket HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
```

A simpler fix was to install `uvicorn[standard]`, but that installed quite a few more things. Just `python-socketio==5.9.0` gave me `simple-websocket==1.0.0` as a side effect and that was enough.

I suppose we could also declare `simple-websocket` ourselves in the `api` dependency block.

joanise:
Warning: you accidentally undid #365. Maybe we need to make the change in pyproject.toml too. That fixes a prod CVE, so we really want to apply it.

dhdaines:
> Warning: you accidentally undid #365. Maybe we need to make the change in pyproject.toml too. That fixes a prod CVE, so we really want to apply it.

Oh, oops.  This may be due to "git rebase"'s merge conflicts being backwards and hard to understand.

dhdaines:
> > Warning: you accidentally undid #365. Maybe we need to make the change in pyproject.toml too. That fixes a prod CVE, so we really want to apply it.
> 
> Oh, oops. This may be due to "git rebase"'s merge conflicts being backwards and hard to understand.

I don't understand, though ... how did the updated gunicorn dependency disappear from the main branch?

joanise:
Alright, finally I can say I agree: we're there! We can merge this.

Thank you for your patience working through all the details, and putting up with my stale environments which, while annoying, bring us to express our actual dependencies more accurately, so I don't actually feel bad about it. 😉

joanise (approved):
🎉

joanise:
> > Oh, oops. This may be due to "git rebase"'s merge conflicts being backwards and hard to understand.
> 
> I don't understand, though ... how did the updated gunicorn dependency disappear from the main branch?

You almost certainly made a mistake in handling a merge conflict.

This is the commit that undoes it:

```
commit cfc50c6ef1
Author: David Huggins-Daines <dhd@ecolingui.ca>
Date:   Thu Apr 4 10:09:02 2024 -0400

    fix: update prod environment and workflow
```

joanise:
In your defense, merge conflict resolution is indeed quite confusing. I'm quite paranoid about them, often doing extensive reviewing of the commit I'm trying to rebase and the changes that were there before, and not just relying on the conflict resolution tools.

PR #361, opened 2024-04-13
Urgent: Heroku does not like dynamic versioning

dhdaines:
For *whatever* reason, Heroku expects to be able to run your app in-place without installing it in any way.  This is not a great idea when your "app" is really a library which also happens to contain an app, but whatever.

Luckily it provides you with hooks (not sure where these are documented but the code says "people rely on this") so we can run `pip install -e .` to fix that... and under certain circumstances (Heroku still believes `python setup.py develop` is a current practice it seems?) it will even do this for you.

Except of course that even though Heroku uses git to deploy, it in no way gives you access to the git repository itself!  **NO HISTORY FOR YOU!**  And thus no version information... So we have to use the same technique as in GitHub CI to ensure not only that `g2p/_version.py` gets created, but that the build doesn't crash.


dhdaines:
UGH. I cannot merge this because it needs approval. Remind me not to do things on Friday evening.

roedoejet (approved):
Thanks for the fix! David. Heroku is a bit cumbersome/idiosyncratic sometimes, yes 🙃

PR #362, opened 2024-04-15
chore: delete deprecated code

joanise:
Now that we've release g2p 2.0, we can simply delete all the code to accept and warn on the deprecated tok_lang argument to make_g2p.

While I'm there, clean up other compatibility validation code.


dhdaines (approved):
Yup!

PR #363, opened 2024-04-15
fix: make the g2p library tests still run on Python 3.7

joanise:
All production API code, and API testing code, requires Python 3.7, but the base g2p CLI and library module still support Python 3.7 and need to be testing in CI.

@dhdaines This PR on top of #360 would fix my concerns regarding exercising g2p library tests with Python 3.7


dhdaines (approved):
rebased, merging!

PR #368, opened 2024-04-17
ci: make the PR comment optional so dependabot PRs pass CI

joanise:
Dependabot does not have permission to post PR comments, but that should not make it fail CI.


dhdaines (approved):
I don't really know what this does but it seems sensible

PR #369, opened 2024-04-18
fix(build): depend on gitlint-core, not gitlint

joanise:
From https://jorisroovers.com/gitlint/latest/installation/:
> By default, gitlint is installed with pinned dependencies. To install
> gitlint with looser dependency requirements, only install gitlint-core

We don't want gitlint's pinned requirements to cause trouble, which they did when I tried to install ReadAlongs/Studio in a sandbox where the latest g2p was already installed with .[dev].

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,7 @@ dev = [
   "g2p[test]",
   "black>=23; python_version < '3.8'",
   "black~=24.3; python_version >= '3.8'",
-  "gitlint>=0.17.0",
+  "gitlint-core>=0.17.0",
   "isort>=5.10.1",
   "mypy>=1; python_version < '3.8'",
   "mypy>=1.8.0; python_version >= '3.8'",


joanise:
This PR is growing my more stuff I discover as I try to clean it up...

1) it's been annoying me how PRs that don't touch code often lose a little bit a coverage. I think I found out why, and added some tests in `test_studio.py` to fix that.

2) mypy complained about a bunch of stuff. typing extensions: `--install-types` fixes that, added it to `.pre-commit-config.yaml` (it would be nice to have the pre-commit mypy see the pyproject.toml tool.hatch.envs.dev.scripts options, but I'm not sure that's realistic. -- Ah-ha! Found it: added `install_types` and `non_interactive` to `[tool.mypy]` in `pyproject.toml`.

g2p/mappings/utils.py
```diff
@@ -86,7 +86,8 @@ def expand_abbreviations(data: str, abbs: Dict[str, List[str]], recursion_depth=
     """
     if recursion_depth > 10:
         raise exceptions.RecursionError(
-            "Too many levels of recursion in your abbreviation expansion. Check your abbreviations for circular references."
+            "Too many levels of recursion in your abbreviation expansion. "
```
joanise, line 89: This and several other reformatting changes are just to appease flake8

g2p/mappings/utils.py
```diff
@@ -751,9 +758,7 @@ def check_mapping_types(self) -> "_MappingModelDefinition":
             and self.rules_path is None
         ):
             LOGGER.warning(
-                exceptions.MalformedMapping(
-                    "You have to either specify some rules or a path to a file containing rules."
-                )
+                "Empty mapping: specify some rules or a path to a file containing rules."
```
joanise, line 761: The exception we used to use here printed a lot of irrelevant information, instead, here is a narrow message with just the relevant details.

g2p/mappings/utils.py
```diff
@@ -808,8 +814,8 @@ def validate_preserve_case(self):
     def add_parent_dir(cls, value: Any, info: ValidationInfo):
         """If there is a parent directory, prepend it to all path fields."""
         if isinstance(value, (str, Path)):
-            if info.data.get("parent_dir", None):
-                value = Path(info.data["parent_dir"]) / value
+            if info.data.get("parent_dir", None):  # type: ignore[attr-defined]
+                value = Path(info.data["parent_dir"]) / value  # type: ignore[attr-defined]
```
joanise, line 818: I tried to make mypy see this as correct, as it is valid Pydantic, but I found no way to make it stop saying that ValidationInfo has no data field, try as a might.

dhdaines:
> From https://jorisroovers.com/gitlint/latest/installation/:
> 
> > By default, gitlint is installed with pinned dependencies. To install
> > gitlint with looser dependency requirements, only install gitlint-core
> 
> We don't want gitlint's pinned requirements to cause trouble, which they did when I tried to install ReadAlongs/Studio in a sandbox where the latest g2p was already installed with .[dev].

oh oops! my fault!

dhdaines:
> 2. mypy complained about a bunch of stuff. typing extensions: `--install-types` fixes that, added it to `.pre-commit-config.yaml` (it would be nice to have the pre-commit mypy see the pyproject.toml tool.hatch.envs.dev.scripts options, but I'm not sure that's realistic. -- Ah-ha! Found it: added `install_types` and `non_interactive` to `[tool.mypy]` in `pyproject.toml`.

Ah good!  For a while there I had tried running `mypy --install-types --non-interactive` when creating the dev environment but this is a bad idea since it will actually run the mypy checks (and there is no way to disable this) so if you have any errors, creating the environment will fail.

g2p/mappings/utils.py
dhdaines: hm. have we enabled the [Pydantic mypy plugin](https://docs.pydantic.dev/latest/integrations/mypy/#enabling-the-plugin)?  I thought we did but maybe we didn't yet

g2p/mappings/utils.py
dhdaines: Note that although black will reformat things at "80-ish" columns we have lots and lots and lots of untouched legacy code with way more columns than that so I had to set the limit at 120.  Not sure if we want to go to the [recommended flake8 options for black](https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#flake8) or not.

g2p/mappings/utils.py
dhdaines: hm, yes, it is enabled...

dhdaines: what version of mypy are you using? I don't get an error here with 1.9.0

dhdaines (approved):
This all looks good!  Sorry for the spam earlier, I should have put it all in a review.

g2p/mappings/utils.py
joanise: This was our old flake8 config in `setup.cfg` before your hatch PR, probably just want to go back to these settings?
```
[flake8]
ignore = E203, E302, E501, E999, W503
max-line-length = 88
max-complexity = 12
select = B,C,E,F,W,B9
```

joanise: Notice E501 there, in particular. The fact we've been silencing it would explain why we have many screens of it that we've ignored so far.

g2p/mappings/utils.py
joanise: Oh, good question! I have 1.8.0 in one env, I'll test to see if 1.9.0 fixes it.

joanise: Nope, mypy 1.9.0, Python 3.8.17, I still get there error.

g2p/mappings/utils.py
dhdaines: oh... yeah... well if you ignore E501 then max-line-length does nothing :)

g2p/mappings/utils.py
joanise: Ah, OK, `pydantic>=2.4.0` fixes it. 2.3.0 has the warning.

PR #370, opened 2024-04-18
A couple of Studio fixes

dhdaines:
One rather important one - we definitely *do not* need to send CORS from the Studio app to accept all origins, this was a (bad) leftover from some example code.  Studio is definitely same-origin and should always be.

One less important one - if you delete all the input, the last character sticks around.  It isn't a problem to request conversion of an empty string so we should allow it to happen.

Also add the G2P_STUDIO_DEBUG environment variable which prints out all the socket.io messages, very useful for debugging.

### DIFF
--- a/g2p/static/custom.js
+++ b/g2p/static/custom.js
@@ -437,23 +437,23 @@ var trackIndex = function() {
 
 var convert = function() {
     // prevent conversion from happening before TABLES, ABBS, and SETTINGS are populated.
-    if (TABLES.length > 0 && ABBS.length > 0) {
-        let index = trackIndex()
-        var input_string = $('#input').val();
-        if (index) {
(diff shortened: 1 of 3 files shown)


g2p/static/custom.js
```diff
@@ -437,23 +437,23 @@ var trackIndex = function() {
...
-            let mappings = getIncludedMappings()
-            conversionSocket.emit('conversion event', {
-                data: {
-                    index,
-                    input_string,
-                    mappings
-                }
-            });
+    if (TABLES.length == 0 || ABBS.length == 0)
+        return;
```
joanise, line 441: Why did we require a non-empty ABBS? That makes no sense to me.

joanise (approved):
Looks good, I just have a question about the original code, not your changes.

PR #371, opened 2024-04-19
feat: Salish APA mapping

joanise:

- [x] `config-g2p.yaml` file includes all author names, and settings necessary
- [x] You agree to license your contribution under the same license as this project (see [LICENSE](https://github.com/roedoejet/g2p/blob/main/LICENSE) file).

* **Other information**:


g2p/mappings/langs/sal/sal_equiv.csv
```diff
@@ -0,0 +1,4 @@
+’,'
+‘,'
+\u0315,'
+\u0301{1}\u0323{2},\u0323{2}\u0301{1}
```
roedoejet, line 4: what about other single quote lookalikes? I wonder if we should have something generic for this, basically all Unicode confusables:

```
｀ ΄ ＇ ˈ ˊ ᑊ ˋ ꞌ ᛌ 𖽒 𖽑 ‘ ’ י ՚ ‛ ՝ ` ' ′ ׳ ´ ʹ ˴ ߴ ‵ ߵ ʻ ʼ ᾽ ʽ ῾ ʾ ᾿
```

roedoejet (approved):
Looks great @joanise ! Congrats to you and Tony. I wonder if we should incorporate some sort of standard single-quote normalization based on Unicode confusables?

g2p/mappings/langs/sal/sal_equiv.csv
joanise: I've given a lot of thought to this question for every mapping I touched using these for ejectives, glottal stops, stress markers, etc, and I've always stepped back from such a general solution, because the correct interpretation depends on the language, and I worry a global solution will be wrong too often.

For this mapping it's a very partial solution based on data I've seen. I almost certainly missed some real-life use cases, but until I see them I'm not convinced my assumptions on how to handle them will be correct. But maybe I'm being too conservative, I don't know.

PR #376, opened 2024-04-29
fix: normalize ó in mohawk

MENGZHEGENG:
A minor change where normalizing ó to accommodate different typing inputs


roedoejet (changes requested):
Thanks for finding this Tim! We need to run `g2p update` and include some tests. I'll add that in a second here.

PR #381, opened 2024-05-22
feat: add logic to auto-delete as_is support in g2p 3

### LINKED ISSUE #374: Mapping.as_is has been deprecated for a long time, we should be able to delete it
The documentation `Mapping.as_is` has been deprecated since 0.6, I think by now it should be safe to delete it from the code.

Does that require a "breaking change" and therefore a major version bump? If so, meh, that would be annoying. 

Maybe a conditional fatal error, like we had for something else that got deprecated a while ago and removed in 2.0:
```
if v < 3:
   warn as_is is deprecated and will be removed from g2p 3
else:
   error out with similar error
```
and then we won't forget to remove it in v3...

joanise:
### PR Goal? 

`as_is` has been deprecated for a long time, but we didn't declare when it would be removed. Let this be that declaration.

### Fixes? 

Fixes #374

### Feedback sought? 

sanity check

### Priority? 

low

### Tests added? 

was already tested

### How to test? 

If you want to see it fail, change `version_tuple < (3,):` to `< (2,)`, that will make the as_is unit test fail now, which should get deleted when we release 3.

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/g2p/mappings/__init__.py
+++ b/g2p/mappings/__init__.py
@@ -16,6 +16,7 @@ import yaml
 from pydantic import BaseModel
 
 from g2p import exceptions
+from g2p._version import version_tuple
 from g2p.log import LOGGER
 from g2p.mappings.langs import _LANGS, _MAPPINGS_AVAILABLE
 from g2p.mappings.langs import __file__ as LANGS_FILE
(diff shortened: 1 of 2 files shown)


roedoejet (approved):
Fair enough - nice to actually deprecate some things that have been deprecated for a while.

PR #384, opened 2024-07-03
Fix: allow running the app and tests from any directory

joanise:
Two related fixed enabling running `app.py` from directories other than the root:

1. fix(app): do not rely on running at the g2p root dir

Use g2p.__file__ to locate g2p/static and g2p/templates, don't rely on finding them under the current working directory.

This allows us to run `uvicorn g2p.app:APP --reload --port 5000` and the gunicorn production variant from anywhere, not just the g2p sandbox root.

This also allows us to run `cd g2p; tests/run.py` or `cd g2p/tests; ./run.py`, or `./test_api_resources.py`, etc.

2. refactor: rename types->shared_types to avoid collision

Running app.py from g2p/g2p, libraries load g2p.types instead of standard types and thus app.py was broken.


roedoejet (commented):
lgtm, thanks @joanise !

roedoejet (approved):
lgtm, thanks @joanise !

PR #385, opened 2024-07-05
Miscenalleous cli improvement for convert

joanise:
### PR Goal? 

misc small improvements:
 - accept `-` and `/dev/stdin` as stdin (the latter on Linux only)
 - when reading a file or stdin, read it and process it line by line - that was just being lazy when I slurped the file up front
 - make some unit tests more quiet

### Feedback sought? 

standard code review - use SemanticDiff to make it easier to read, though:
https://app.semanticdiff.com/gh/roedoejet/g2p/pull/385

### Priority? 

low

### Tests added? 

yup

### How to test? 

 - `g2p convert my_file.txt fra fra-ipa` should continue to work as expected, processing the contents of `my_file.txt`
 - `echo blah blah | g2p convert - fra fra-ipa` should now work (real use case: replace echo by a meaningful process)
 - on Linux only: `g2p convert <(echo blah blah) fra fra-ipa` now also works

The last two cases came up as I was preparing and processing files for EveryVoice, where I wanted to cut a column out of a psv file and pass it to g2p without having the same temporary files on disk. That's what triggered this patch in the first place.

### Confidence? 

high

### Version change? 

Possibly a minor since we're adding a feature, but really I think of this as just a patch: this is how it should have worked in the first place.

We've done a bunch of changes since 2.0.0, though, so we're probably due for releasing 2.1.0, especially for api/v2.

### DIFF
--- a/g2p/cli.py
+++ b/g2p/cli.py
@@ -442,6 +442,13 @@ def generate_mapping(  # noqa: C901
     help="Show all the conversion steps applied.",
 )
 @click.option(
+    "--file",
+    "-f",
+    default=False,
+    is_flag=True,
+    help="Read input from plain-text file INPUT_TEXT.",
+)
+@click.option(
     "--check/--no-check",
     "-c",
     default=False,
@@ -478,6 +485,7 @@ def convert(  # noqa: C901
     input_text,
     tok,
     check,
+    file,
     debugger,
     pretty_edges,
     tok_lang,
@@ -531,20 +539,23 @@ def convert(  # noqa: C901
         raise click.UsageError(
             f"Path between '{in_lang}' and '{out_lang}' does not exist"
         )
-    # Figure if the input text is on the command line or in a file
-    input_text_is_a_file = (
-        os.path.exists(input_text)
-        and input_text.endswith("txt")
-        or re.match(r"/dev/(fd/[0-9]*|stdin)", input_text)
-    )
     to_close = None
     try:
-        if input_text_is_a_file:
-            to_close = lines = open(input_text, encoding="utf8")
-        elif input_text == "-":
-            lines = sys.stdin
+        if file:
+            if input_text == "-":
+                lines = sys.stdin
+            else:
+                try:
+                    to_close = lines = open(input_text, encoding="utf8")
+                except FileNotFoundError as e:
+                    raise click.UsageError(f"Could not open file {input_text}: {e}")
         else:
             lines = [input_text]
+            if os.path.exists(input_text) and input_text.endswith(".txt"):
+                LOGGER.warning(
+                    "The old heuristic for detecting file input is deprecated, "
+                    f"specify --file if you meant to read your input text from file {input_text}."
+                )
         # Determine which tokenizer to use, if any
         if tok is not None and not tok and tok_lang is not None:
             raise click.UsageError(
@@ -572,7 +583,7 @@ def convert(  # noqa: C901
             if len(outputs) > 1:
                 click.echo(pprint.pformat(outputs, indent=4))
             else:
-                click.echo(tg.output_string, nl=not input_text_is_a_file)
+                click.echo(tg.output_string, nl=not file)
     finally:
         if to_close is not None:
             to_close.close()
--- a/g2p/cli.py
+++ b/g2p/cli.py
@@ -537,39 +537,45 @@ def convert(  # noqa: C901
         and input_text.endswith("txt")
         or re.match(r"/dev/(fd/[0-9]*|stdin)", input_text)
     )
-    if input_text_is_a_file:
-        with open(input_text, encoding="utf8") as f:
-            lines = f.readlines()
-    elif input_text == "-":
-        lines = sys.stdin.readlines()
-    else:
-        lines = [input_text]
-    # Determine which tokenizer to use, if any
-    if tok is not None and not tok and tok_lang is not None:
-        raise click.UsageError("Specified conflicting --no-tok and --tok-lang options.")
-    if tok is None:
-        tok = True  # Tokenize by default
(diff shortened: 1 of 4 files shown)


g2p/cli.py
```diff
@@ -532,38 +532,50 @@ def convert(  # noqa: C901
...
-            outputs += [tg.substring_alignments()]
-        if pretty_edges:
-            outputs += [tg.pretty_edges()]
-        if debugger:
-            outputs += [tg.edges, tg.debugger]
-        if len(outputs) > 1:
-            click.echo(pprint.pformat(outputs, indent=4))
+    to_close = None
+    try:
+        if input_text_is_a_file:
```
dhdaines, line 542: Can we just use click.File for this? https://click.palletsprojects.com/en/8.1.x/arguments/#file-arguments

dhdaines (changes requested):
Good idea, but I think we don't need to implement these checks ourselves!

g2p/cli.py
joanise: I don't know if click supports "string OR file" which is what we have here. The original was just a string, then for convenience if the string happened to name a file I'd open and read it, now there are users of this at-first-hidden feature and I'm trying to improve it.
click.File would be a good choice if we weren't abusing the syntax like that.
I suppose this mean suggest changing the interface to have --input-file and --input-text instead of conflating the two, but I'm not keen on making such a breaking change without a strong compelling case (stronger than reusing library code).

dhdaines: oh... "string or file"? so, if it is a string that doesn't name an existing file then we treat it as a sring, otherwise we open it as a file?  This is very strange behaviour for a CLI :)

joanise: Indeed, I rather regret that weird CLI choice indeed, especially with what you see here, but now that some people rely on it, I'm not that keen on changing it. Although it's only a feature you'd use interactively, so it wouldn't be that bad to break it: it would not be a breaking API or breaking library change, so no dependent code would get broken by it, only the people using the feature would have to learn the modified behaviour.

And "string or file" is a little narrower than that: it has to name a file that has a .txt extension and exists, which was initially just a hidden feature for my private use that some users at UBC stumbled upon because it was the obvious thing to do and it happened to work as they expected.

The annoying thing now is that this "string or file" is a mandatory positional argument, and if I were to replace the file variant with `--in FILENAME` or something like, then suddenly the first positional argument would be the optional one, and I really don't know how to implement that in click.

If we change the interface for accepting a filename, I would issue a warning when the old automatic logic would have been triggered documenting the new way to call it.

But anyway, my question now is, do you have a good suggestion for a better CLI behaviour that would be intuitive, easy to implement, and most importantly not change the original behaviour of `g2p convert INPUT_TEXT IN_LANG OUT_LANG` which has been around a long time and would, in my opinion, be a major breaking change if we modified it?

I suppose we could add a `--text-is-filename|--file|-f` Boolean flag, but I kinda prefer the current weird auto-detect behaviour... However with such a flag, we'd still have "string or file", just with simpler triggering logic.

joanise:
@roedoejet @dhdaines This PR is ready to (re)review, with the `--file` option we agreed on yesterday for `g2p convert` reading from a file.

Includes documentation, and an unrelated change to CI I did months ago but that got lost in a PR that never got merged.

Plus making `test_update_schema()` quiet except when there are errors, and making those errors more helpful when the are some.

g2p/cli.py
```diff
@@ -531,39 +539,54 @@ def convert(  # noqa: C901
...
-        if len(outputs) > 1:
-            click.echo(pprint.pformat(outputs, indent=4))
+    to_close = None
+    try:
+        if file:
+            if input_text == "-":
+                lines = sys.stdin
+            else:
+                try:
+                    to_close = lines = open(input_text, encoding="utf8")
```
roedoejet, line 549: why not use `with open(input_text)`? just to prevent the indenting? I prefer using `with` just in case something happens later that prevents `to_close` from actually being closed.

g2p/cli.py
```diff
@@ -531,39 +539,54 @@ def convert(  # noqa: C901
...
+                )
+        # Determine which tokenizer to use, if any
+        if tok is not None and not tok and tok_lang is not None:
+            raise click.UsageError(
+                "Specified conflicting --no-tok and --tok-lang options."
+            )
+        if tok is None:
+            tok = True  # Tokenize by default
+        custom_tokenizer = make_tokenizer(tok_lang) if tok_lang else None
+        # Transduce!!!
```
roedoejet, line 567: lol yes. this is the right number of exclamation marks.

g2p/cli.py
```diff
@@ -531,39 +539,54 @@ def convert(  # noqa: C901
...
+        assert in_lang and out_lang
+        transducer = make_g2p(
+            in_lang, out_lang, tokenize=tok, custom_tokenizer=custom_tokenizer
+        )
+        for line in lines:
+            tg = transducer(line)
+            if check:
+                transducer.check(tg, display_warnings=True)
+            outputs = [tg.output_string]
```
roedoejet, line 576: assuming this is copy/pasted from the old place, so not checking carefully.

roedoejet (approved):
thanks for bringing this up the other day - lgtm @joanise !

g2p/cli.py
joanise: Because if the input is a text string, you're not going to be inside that `with` block.

joanise: I always use `with`, it has to be impossible to use it before I resolve to using something else.

g2p/cli.py
joanise: Yeah, if you hide whitespace changes you'll see this is old code, just shifted because of the new `try:` block.

dhdaines (approved):
Looks good!

Issue #386, opened 2024-08-09
Failing on python 3.8


roedoejet:
We're currently failing on python 3.8 due to a bug that has a PR here: https://github.com/dmort27/panphon/pull/57

dhdaines:
In fact it seems that panphon 0.21.x leads to different mappings being generated:
```
======================================================================                                                                                
FAIL: test_mapping (g2p.tests.test_fallback.FallbackTest)                                                                                             
----------------------------------------------------------------------                                                                                
Traceback (most recent call last):                                                                                                                    
  File "/home/dhd/work/ReadAlongs/g2p/g2p/tests/test_fallback.py", line 75, in test_mapping                                                           
    self.assertEqual(                                                                                                                                 
AssertionError: Lists differ: [Rule[193 chars]put='i', context_before='', context_after='', [593 chars]one)] != [Rule[193 chars]put='ɑ', context_befor
e='', context_after='', [593 chars]one)]                                                                                                              
                                                                                                                                                      
First differing element 1:                                                                                                                            
Rule([24 chars]put='i', context_before='', context_after='', [86 chars]None)                                                                          
Rule([24 chars]put='ɑ', context_before='', context_after='', [86 chars]None)                                                                          
                                                                                                                                                      
  [Rule(rule_input='æ', rule_output='ɑ', context_before='', context_after='', prevent_feeding=False, match_pattern=re.compile('æ'), intermediate_form=
None, comment=None),                                                                                                                                  
-  Rule(rule_input='ɐ', rule_output='i', context_before='', context_after='', prevent_feeding=False, match_pattern=re.compile('ɐ'), intermediate_form=
None, comment=None),                                                                                                                                  
?                                    ^                                                                                                                
                                                                                                                                                      
+  Rule(rule_input='ɐ', rule_output='ɑ', context_before='', context_after='', prevent_feeding=False, match_pattern=re.compile('ɐ'), intermediate_form=
None, comment=None),                                                                                                                                  
?                                    ^                                                                                                                
                                                                                                                                                      
   Rule(rule_input='ɑ̃', rule_output='ɑ', context_before='', context_after='', prevent_feeding=False, match_pattern=re.compile('ɑ̃'), intermediate_form=
None, comment=None),                                                                                                                                  
   Rule(rule_input='β', rule_output='s', context_before='', context_after='', prevent_feeding=False, match_pattern=re.compile('β'), intermediate_form=
None, comment=None),                                                                                                                                  
[...]

PR #387, opened 2024-08-16
fix(deps): pin panphon to 0.19-0.20 as 0.21 breaks many things

dhdaines:
It gives different output and also causes Python 3.8 to crash, as detailed in https://github.com/dmort27/panphon/pull/57

Fixes: #386

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
     "coloredlogs>=15.0.1",
     "networkx>=2.6",
     "openpyxl",
-    "panphon>=0.19",
+    "panphon>=0.19,<0.21",
     "pydantic>=2.4",
     "pyyaml>=5.2",
     "regex",


roedoejet (approved):
this is what I was trying to do with the python 3.8 specific statement since I think python 3.9+ works with panphon >= 0.21.0. Did you ever figure out how to do that? In any case, I'm fine to just pin the whole thing to < 0.21

dhdaines:
> this is what I was trying to do with the python 3.8 specific statement since I think python 3.9+ works with panphon >= 0.21.0. Did you ever figure out how to do that? In any case, I'm fine to just pin the whole thing to < 0.21

Yes, I had that working too, but then I discovered the other problem with panphon 0.21... It looked like this:
```toml
dependencies = [
    "panphon<0.21; python_version<'3.9'",
    "panphon>=0.19; python_version>='3.9'",
```

PR #390, opened 2024-08-22
fix(ci): remove stale job dependency in pythonpublish workflow

joanise:
### PR Goal? 

Fix the python publish workflow so we can publish 2.1.0

### Fixes? 

Bug in the workflow that made yesterday's attempt at 2.1.0 fail.

### Feedback sought? 

sanity

### Priority? 

high

### Tests added? 

nope

### How to test? 

Would require forking and removing the real work from this, so not easily. I'm confident in the patch, but if it breaks again we'll just have to try once more. At least we know nothing goes to PyPI unless everything else succeeds.

### Confidence? 

High.

### Version change? 

yup...

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -45,7 +45,6 @@ jobs:
       - tests
       - matrix-tests
       - build
-      - build-docs
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


roedoejet (approved):
lgtm thanks @joanise

PR #393, opened 2024-08-26
Revamp schema versioning and dev version trigger

### LINKED ISSUE #392: Revamp the schema update tests
Currently, we unit test that there is a schema file matching the current major.minor. This is insufficient in a few ways:
 - if we changed the models (and therefore the schemas) from what is saved in `g2p/mappings/.schema`, that should be an error (indeed, we missed the fact that the 2.0 schema was missing two fields when we published g2p 2.0.0)
 - if we're working on the next minor or major version, hatch will still think the major.minor is the previous one (from the tags) with a patch bump, so we're not going to check the right .schema file version
 - when we bump the minor but no changes are required to the schema, we should not publish a new schema: the test should be that if the most recent schema already generated matches the current model (i.e., the schema we would generate), that's a pass.
 - we should also distinguish between a published schema file and an unpublished one: if it's already on the schema store, it's locked and we should never change it. If it's not published yet, changing it as the model changes is what we need to do. In a perfect world unit testing would enforce these rules.

joanise:
### PR Goal? 

Clean up how schemas a generated and unit tested

### Fixes? 

Fixes #392 

Also fixes the chicken and egg problem when we start working on the next minor: since the `setuptools_scm` uses tags to determine the current version, you'd have to tag 2.2.0 if the current is 2.1.0 before it would install with the right version, but that's not right since 2.2.0 would not have been published yet. But, in fact, the solution is simple:
 - when we start 2.2.0, tag 2.2.0.dev0
 - prevent pythonpublish.yml from actually publishing that to PyPI, since it's just dev (this PR does that with 65499d6)
 - so ongoing work will now except the schema at 2.2, but the current published version will still be 2.1.x

### Feedback sought? 

general validation

@roedoejet make sure the schema logic is good
@dhdaines confirm my interpretation of dev tagging with dynamic versioning

### Priority? 

low

### Tests added? 

yup, loads!

### How to test? 

Run `g2p update-schema` with the current schema being up to date, wrong, a new version without change, a new version with change (unit testing covers all these cases).

### Confidence? 

moderate

### Version change? 

No, but will make the next change cleaner.


joanise:
I added one more thing: make the SCM pretend version thing DRY, and validate it so it stays in sync.

.github/workflows/tests.yml
```diff
@@ -9,7 +9,6 @@ on:
 # This **must** match the Major.Minor version of the JSON schema file
```
dhdaines, line 9: probably should remove this comment :)

joanise: Right, yes! :)

joanise:
CI was failing because pydantic 2.9 changes the schema. :(
Fixed by excluding it, but that's annoying because we're generally compatible with it, it's just an irrelevant change, but which has visible effect on the schema file.

g2p/cli.py
```diff
@@ -700,34 +701,64 @@ def update(in_dir, out_dir):
...
+    if schema_path.exists():
+        with open(schema_path, encoding="utf8") as f:
+            old_schema = json.load(f)
+        if old_schema == json_schema:
+            print(f"Schema {schema_path} is already up to date.")
+        else:
+            print(
+                dedent(
+                    f"""
+                    Schema {schema_path}
```
dhdaines, line 736: black isn't complaining about these line lengths?

g2p/tests/test_cli.py
```diff
@@ -96,31 +120,54 @@ def test_update(self):
...
-            (
-                Path(__file__).parent.parent
-                / "mappings"
-                / ".schema"
-                / f"g2p-config-schema-{MAJOR_MINOR_VERSION}.json"
-            ).exists()
-        )
-
     def test_update_schema(self):
+        # It's an error for the currently saved schema to be out of date
```
dhdaines, line 124: I'll just assume all of this works!

dhdaines (approved):
Ugh, I am not very awake, I clickd "add single comment" when I meant to click "start a review.  Sorry about that.  This all looks good to me, assuming it passes tests and coverage.

.github/workflows/matrix-tests.yml
```diff
@@ -21,8 +21,9 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
+        shell: bash
         run: |
           python -m pip install --upgrade pip
-          pip install -e .[test]
+          SETUPTOOLS_SCM_PRETEND_VERSION=`cat .SETUPTOOLS_SCM_PRETEND_VERSION` pip install -e .[test]
```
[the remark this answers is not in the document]
joanise: I was going to use `$(...)`, but I decided backticks were a better choice. Now I don't remember why...
Does `$(...)` work with POSIX Bourne shell? I might have hesitated because of not being sure that would be the case, but maybe that's misguided.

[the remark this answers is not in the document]
joanise: OK, no, that was misguided, https://stackoverflow.com/questions/22709371/backticks-vs-braces-in-bash says `$()` is POSIX compliant, and backticks are deprecated! I learned something today.

g2p/tests/test_cli.py
joanise: It works too well, killed CI when it ran with Pydantic 2.9.1... :(

g2p/cli.py
joanise: Adjusted a bit, but they're sized to look decent after dedenting, so they're going to stay somewhat long.
I agree they were too long, though, so I adjusted them somewhat.

.github/workflows/matrix-tests.yml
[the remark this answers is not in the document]
dhdaines: yeah the nice thing about `$()` is that it nests, e.g.
```sh
echo $(echo $(echo foo) $(echo bar))
```

PR #396, opened 2024-09-11
feat: allow panphon 0.21 where possible

### LINKED ISSUE #388: panphon >= 0.21 may give better mappings
Once it works with Python 3.8 we should unpin it and redo the test mappings.  This error isn't really an error, since I think we all agree that 'ɑ' is a much better mapping for 'ɐ' than 'i'
```
-  Rule(rule_input='ɐ', rule_output='i', context_before='', context_after='', prevent_feeding=False, match_pattern=re.compile('ɐ'), intermediate_form=
None, comment=None),                                                                                                                                  
?                                    ^                                                                                                                
                                                                                                                                                      
+  Rule(rule_input='ɐ', rule_output='ɑ', context_before='', context_after='', prevent_feeding=False, match_pattern=re.compile('ɐ'), intermediate_form=
None, comment=None),                                                                                                                                  
```

joanise:
And make the unit test accept 0.21 and older versions.

Fixes #388

tested on Windows with Python 3.8 and 3.10, I get panphon 0.20 in both cases.
Tested on Linux with Python 3.10, I get panphon 0.21.2.

The unit tests pass, with a small patch fixing the error documented in #388 but in a way that tests still pass with the older panphon.


joanise:
Validated with matrix testing that we get panphon 0.21 where expected.
See https://github.com/roedoejet/g2p/actions/runs/10813025674

dhdaines (approved):
This seems good with the slight caveat that ... the mappings produced by panphon 0.21 really aren't better (matching ɐ to i is ... weird, particularly if it ends up as IY in arpabet)

PR #397, opened 2024-09-11
Refactor: replace networkx by a very lightweight custom DiGraph class

### LINKED ISSUE #394: RAM usage: replace networkx elements used by the web API server
g2p is used on the ReadAlong Studio back-end server, and causes a fairly high amount of RAM to get used.

 - importing `networkx` (any subpart of it, even just the top-level import) costs about 30MB
 - importing the languages database costs about 60MB, 40 or 50 of which is likely the English lexicon (see #395)

If we can rewrite the tiny part of `networkx` we use ourselves, or maybe use `networkx` in general except for the parts `readalongs/web_api.py` needs, that would be nice -- basically `g2p update` would use `networkx` but `make_g2p` and loading the langs database would not. The algorithms we use from networkx are quite simple, after all:
 - [ ] has path
 - [ ] shortest path
 - [ ] ancestors
 - [ ] descendants

Given the simplicity of the g2p language DAG, we don't need full Dijkstra, implementing `has_path(src, tgt)` as `tgt in descendants(src)` would not likely have any performance cost, and there are so few cases where the is more than one path between two nodes, implementing `shortest_path(src,tgt)` with a simple BFS from src would be unlikely to be more expensive than a real Dijkstra implementation.

joanise:
### PR Goal? 

Reduce the memory footprint of g2p and RAS API servers, by replacing networkx (~30MB in RAM to import) by the most minimal class that meets the needs of g2p.

### Fixes? 

Fixes #394

### Feedback sought? 

This is pretty core to the library, so careful review of all the code touched by this PR. All testing passes, and I expect coverage to be excellent (yay for our previous test development in this library), so I'm pretty confident it's all good, but still, I'd like careful review.

### Priority? 

medium

### Tests added? 

yup

### How to test? 

 - `run_tests.py` -> see no failures
 - `g2p update` -> see no changes in the generated files
 - `g2p convert asdf fra eng-arpabet` -> see the same output as before
 - all other g2p operations are affected, so play around if you feel the need, but CI covers it all already, so not really necessary.

### Confidence? 

medium-high

### Version change? 

No. I was careful that it's pure refactoring, there are no external behaviour changes.

But this will be worth a patch bump so `pip install g2p` can stop installing `networkx`.

### DIFF
--- a/g2p/mappings/langs/network_lite.py
+++ b/g2p/mappings/langs/network_lite.py
@@ -133,8 +133,9 @@ class DiGraph(Generic[T]):
         visited: Dict[T, Union[T, None]] = {
             u: None
         }  # dict of {node: predecessor on shortest path from u}
-        queue: deque[T] = deque()
-        while True:
+        queue: deque[T] = deque([u])
+        while queue:
+            u = queue.popleft()
             if u == v:
                 rev_path: List[T] = []
                 nextu: Union[T, None] = u
@@ -146,9 +147,7 @@ class DiGraph(Generic[T]):
                 if neighbour not in visited:
                     visited[neighbour] = u
                     queue.append(neighbour)
-            if len(queue) == 0:
-                raise ValueError(f"No path from {u} to {v}")
-            u = queue.popleft()
+        raise ValueError(f"No path from {u} to {v}")
 
 
 NodeDict = TypedDict("NodeDict", {"id": Any})
--- a/g2p/mappings/langs/network_lite.py
+++ b/g2p/mappings/langs/network_lite.py
@@ -1,0 +1,164 @@
+from collections import deque
+from typing import Any, Iterable, Iterator
+
+
+class DiGraph:
+    """A simple directed graph class
+
+    Most functions raise KeyError if called with a node u or v not in the graph.
+    """
+
+    def __init__(self) -> None:
+        """Contructor, empty if no data, else load from data"""
+        self._edges: dict = {}
+
+    def clear(self):
+        """Clear the graph"""
+        self._edges.clear()
+
+    def update(self, edges: Iterable, nodes: Iterable):
+        """Update the graph with new edges and nodes"""
+        for node in nodes:
+            self.add_node(node)
+        for u, v in edges:
+            self.add_edge(u, v)
+
+    def add_node(self, u):
+        """Add a node to the graph"""
+        if u not in self._edges:
+            self._edges[u] = []
+
+    def add_edge(self, u, v):
+        """Add a directed edge from u to v"""
+        self.add_node(u)
+        self.add_node(v)
+        if v not in self._edges[u]:
+            self._edges[u].append(v)
+
+    def add_edges_from(self, edges: Iterable):
+        """Add edges from a list of tuples"""
+        for u, v in edges:
+            self.add_edge(u, v)
+
+    @property  # read-only
+    def nodes(self):
+        """Return the nodes"""
+        return self._edges.keys()
+
+    @property  # read-only
+    def edges(self) -> Iterator:
+        """Iterate over all edges"""
+        for u, neighbours in self._edges.items():
+            for v in neighbours:
+                yield u, v
+
+    def __contains__(self, u) -> bool:
+        """Check if a node is in the graph"""
+        return u in self._edges
+
+    def has_path(self, u, v) -> bool:
+        """Check if there is a path from u to v"""
+        if v not in self._edges:
+            raise KeyError(f"Node {v} not in graph")
+        visited: set = set()
+        return self._has_path(u, v, visited)
+
+    def _has_path(self, u, v, visited: set) -> bool:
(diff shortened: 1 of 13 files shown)


g2p/mappings/langs/network_lite.py
```diff
@@ -0,0 +1,164 @@
...
+        self._edges.clear()
+
+    def update(self, edges: Iterable, nodes: Iterable):
+        """Update the graph with new edges and nodes"""
+        for node in nodes:
+            self.add_node(node)
+        for u, v in edges:
+            self.add_edge(u, v)
+
+    def add_node(self, u):
```
dhdaines, line 26: Should be `u: Hashable` as not all types can actually be keys in a dict (likewise for `add_edge`, etc)

g2p/mappings/langs/network_lite.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def __init__(self) -> None:
+        """Contructor, empty if no data, else load from data"""
+        self._edges: dict = {}
+
+    def clear(self):
+        """Clear the graph"""
+        self._edges.clear()
+
+    def update(self, edges: Iterable, nodes: Iterable):
```
dhdaines, line 19: As below, `Iterable[Hashable]` I think...

g2p/mappings/langs/network_lite.py
```diff
@@ -0,0 +1,164 @@
...
+
+    graph = DiGraph()
+    for node in data["nodes"]:
+        graph.add_node(node["id"])
+    for edge in data["links"]:
+        graph.add_edge(edge["source"], edge["target"])
+    return graph
+
+
+def node_link_data(graph: DiGraph):
```
dhdaines, line 154: Should have a return annotation of `TypedDict` (with the appropriate fields)

g2p/mappings/langs/network_lite.py
```diff
@@ -0,0 +1,164 @@
...
+        Raises:
+            KeyError: if u or v is not in the graph
+            ValueError: if there is no path from u to v
+        """
+
+        if v not in self._edges:
+            raise KeyError(f"Node {v} not in graph")
+        visited = {u: None}  # dict of {node: predecessor on shortest path from u}
+        queue: deque[Any] = deque()
+        while True:
```
dhdaines, line 119: This can just be `while queue`.  If it terminates without finding a path then raise ValueError.  I have written versions of this function about a zillion times, e.g. https://github.com/dhdaines/pdfplumber/blob/develop/pdfplumber/structure.py#L125

g2p/mappings/langs/network_lite.py
```diff
@@ -0,0 +1,164 @@
...
+            for neighbour in self._edges[u]:
+                if neighbour not in visited:
+                    visited[neighbour] = u
+                    queue.append(neighbour)
+            if len(queue) == 0:
+                raise ValueError(f"No path from {u} to {v}")
+            u = queue.popleft()
+
+
+def node_link_graph(data: dict):
```
dhdaines, line 135: I think this still needs a return type annotation even though it's obvious from type inference?  Not sure about that though...

dhdaines (changes requested):
Great!  I have various small changes mostly related to type annotations... should I just add them in a commit?

g2p/mappings/langs/network_lite.py
joanise: I almost wrote it with `while queue`, but I chose to not do `queue.append(u)` before the loop starts, and exit in the second last statement of the loop instead of in the while condition.

It's the same algorithm, I know, but I have a mild preference for the way I wrote it, which avoids pushing `u` and then immediate popping it as the loop starts. Trivial and pointless optimization, I'm aware... but it felt right.

g2p/mappings/langs/network_lite.py
joanise: Yeah,  `-> DiGraph` would be better.

joanise:
> Great! I have various small changes mostly related to type annotations... should I just add them in a commit?

I'll do them as a learning exercise, you're suggesting things I was not familiar with.

joanise:
> I'll do them as a learning exercise, you're suggesting things I was not familiar with.

Never mind, you said in another channel you've started doing the changes, so go for it.

Thanks for the review, it's very helpful!

g2p/mappings/langs/network_lite.py
dhdaines: Ah, usually I do `queue = deque([u])` ... it makes the code shorter...

dhdaines:
> > I'll do them as a learning exercise, you're suggesting things I was not familiar with.
> 
> Never mind, you said in another channel you've started doing the changes, so go for it.
> 
> Thanks for the review, it's very helpful!

Okay! It's done here, with generics, etc:   #398

dhdaines (commented):
Looks good to me!

dhdaines (approved):

PR #398, opened 2024-09-11
Add more types to network_lite

dhdaines:
More types! More types!


joanise:
Wow, that is some serious typing work! Thanks for doing this, I would not have known how to do it all.
Maybe add one more: 
```
LANGS_NETWORK: DiGraph[str] = load_network()
```
in `langs/__init__.py`

dhdaines:
One thing to note here: I'm using (with some difficulty) `TypedDict` to check the types for `node_link_graph` and `node_link_data` but this is actually not very useful because they deal with data that is loaded from JSON at runtime and thus always has type `Any`.  So, in fact, we should use Pydantic here instead ;-)

dhdaines:
> One thing to note here: I'm using (with some difficulty) `TypedDict` to check the types for `node_link_graph` and `node_link_data` but this is actually not very useful because they deal with data that is loaded from JSON at runtime and thus always has type `Any`.

Run-time validation is really a separate issue here (and not strictly necessary given that we know where the data comes from) so I'll leave it for the future.

joanise (approved):
This is awesome, thank you!

PR #399, opened 2024-09-11
Enable type-checking and correct some resulting problems

dhdaines:
A few functions that had type hints were not actually being type-checked, because there weren't any in the function signature.

This revealed a few problems:
- `click.BadParameter`'s type signature doesn't match its documentation (the documentation is correct)
- mypy doesn't narrow strings to literal types properly (see https://github.com/python/mypy/issues/12535)
- we declare `Mapping.rules` as `List[Rule]` but then we put ... other stuff in it, which causes some problems.  Stop doing this and simplify the code so that we don't need a noqa.


g2p/mappings/__init__.py
```diff
@@ -176,34 +178,27 @@ def process_model_specs(self):  # noqa: C901
...
+                    setattr(rule, k, func(value))
+
         non_empty_mappings: List[Rule] = []
         for i, rule in enumerate(self.rules):
-            if isinstance(rule, dict):
-                rule = Rule(**rule)
             # Expand Abbreviations
-            if (
-                self.abbreviations
-                and self.rules
```
dhdaines, line 194: **obviously* self.rules has non-zero length here because we are iterating over it

g2p/mappings/__init__.py
```diff
@@ -176,34 +178,27 @@ def process_model_specs(self):  # noqa: C901
...
-                and "match_pattern" not in self.rules[0]
```
dhdaines, line 195: I really don't know what this was intended to do.  Why only check `self.rules[0]`?  Why not the other rules?

dhdaines: Therefore I removed this check, it seems meaningless - if the intent was not to overwrite a pre-specified `match_pattern` then this is definitely the wrong place to do that.

g2p/mappings/utils.py
```diff
@@ -131,21 +142,22 @@ def normalize(inp: str, norm_form: str):
 
     Also, find any Unicode Escapes & decode 'em!
     """
-    if norm_form not in ["none", "NFC", "NFD", "NFKC", "NFKD"]:
```
dhdaines, line 134: It seems like `None` was a valid value below, but we don't include it here, so that check was meaningless.  Made it meaningful...

g2p/tests/time_panphon.py
```diff
@@ -34,16 +34,16 @@ def getPanphonDistanceSingleton1():
 
 def getPanphonDistanceSingleton2():
     if not hasattr(getPanphonDistanceSingleton2, "value"):
-        setattr(getPanphonDistanceSingleton2, "value", panphon.distance.Distance())
```
dhdaines, line 37: This `setattr` is gratuitous (and flake8 or mypy or somebody warns about it)

g2p/cli.py
```diff
@@ -353,7 +353,7 @@ def generate_mapping(  # noqa: C901
             except MappingMissing as e:
                 raise click.BadParameter(
                     f'Cannot find IPA mapping from "{in_lang}" to "{out_lang}": {e}',
-                    param_hint=["IN_LANG", "OUT_LANG"],
+                    param_hint="'IN_LANG' 'OUT_LANG'",
```
joanise, line 356: Wow, I have no idea how to even trigger this exception!

joanise: Ah, found it. You get it from
`g2p generate-mapping --ipa kwk-boas fra-ipa`
The old code generated
> Error: Invalid value for 'IN_LANG' / 'OUT_LANG': ...

the new code generates
> Error: Invalid value for 'IN_LANG' 'OUT_LANG': ...

I guess the param_hint typing lies, and is in fact more permission that it declares. I kinda like the slash, or even better would be `"'IN_LANG' & 'OUT_LANG'"` because it's the pair of values together that's invalid, not the individual values.

But that's just nitpicking, I know.

g2p/mappings/__init__.py
joanise: I like most of this PR, but I'll have to come back and analyse this change more carefully, because I'm not clear on the original purpose of this check and therefore I'm not confident it's unnecessary. When we wrote it, there must have been a reason, I'd like to dig and find it.

joanise (commented):
This is great stuff. Leave some comments now, I'll have to come back for further analysis, though.

g2p/cli.py
dhdaines: oh! thanks, I actually don't know how it's supposed to be formatted.  It's the type signature that's wrong, not the documentation... so we could restore the old code and just put a 'type: ignore' on it and file a bug against click.

g2p/mappings/__init__.py
dhdaines: yes, I admit I didn't look at `git blame` here, it might reveal the original purpose.  not overwriting an explicit `match_pattern` is something we might want to do, for instance...

dhdaines: This is the commit that added the check for "match_pattern" in the first rule: https://github.com/roedoejet/g2p/commit/8ec4f271b31f21b774619aad418266f6ee85264b @roedoejet do you remember what the intent was?

dhdaines: Yeah, I think it is actually impossible for this to happen now, unless `process_model_specs`, which is called during construction of a `Mapping`, gets called explicitly later on.  Probably what we want to actually do is ensure that nobody does that.

g2p/cli.py
dhdaines: Okay, issue filed with click (linking it here): https://github.com/pallets/click/issues/2777

g2p/mappings/__init__.py
dhdaines: Ok - the way that makes the most sense to do this is to just declare that seeing `match_pattern` or `intermediate_form` in a rule (any rule) is a programmer error, since they are excluded from serialization, and if you specify them, they will get overwritten anyway, so you shouldn't do that.  Added a commit for that!

g2p/mappings/__init__.py
```diff
@@ -176,34 +180,36 @@ def process_model_specs(self):  # noqa: C901
             self.rules = sorted(
                 # Temporarily normalize to NFD for heuristic sorting of NFC-defined rules
                 self.rules,
-                key=lambda x: (
-                    len(normalize(strip_index_notation(x.rule_input), "NFD"))
-                    if isinstance(x, Rule)
-                    else len(normalize(x["in"], "NFD"))
-                ),
```
joanise, line 183: Interestingly, removing this conditional also fixed a bug that has `umista_equiv.csv` sorted before removing its index notation instead of after.
Presumably the real different here is that you're creating `Rule` objects right away all the time, rather initializing the `Mapping` class with either/or.

dhdaines: oh - hopefully I haven't broken some subtle behaviour that we depended on, it seemed to make sense to me to create the `Rule` objects right away, so that they always have the same validation

joanise:
I rebased this PR to the tip of main, and squashed some commits.

I also removed the description from the generated fields, because those description trigger a schema update, which would require a minor version bump and a new schema published to the schema store, and I really don't want to do that just for two hidden fields we're not supposed to populate.

joanise (approved):
Great work, thanks for taking care of these refinements.

Issue #395, opened 2024-09-10
RAM usage: find a way to memory map the English lexicon


joanise:
g2p is used on the ReadAlong Studio back-end server, and causes a fairly high amount of RAM to get used.

 - importing `networkx` (any subpart of it, even just the top-level import) costs about 30MB (see #394)
 - importing the languages database costs about 60MB, 40 or 50 of which is likely the English lexicon

This is more complex than replacing networkx algorithms: if we could find a way to memory map the English lexicon from disk, instead of loading in into RAM, that would let us realize a significant memory saving on the API server, for both the g2p backend and the RAS web-api back-end.

dhdaines:
Ah, I see that I commented on #400 that the representation used there could very easily be memory mapped.

So... this might be pretty easy.  I'll take a look.

dhdaines:
So... in order for memory mapping the representation to make sense we fundamentally need to do two things:

1. Pack it into a single block of memory (as `bytes`, not `str`)
2. Put this block of data in a separate file (presumably, one file for each mapping that uses alignments, though there's only one at the moment) referenced from `langs.json.gz`

What makes this difficult is that the current implementation relies on being able to call `bisect_left` on (blocks of) alignment entries which is definitely not going to work with a packed memory representation, unless we unpack it into a list, which defeats the whole point of doing this.  So we have to:

1. Store the indices of entries in the blob of data separately (we don't want to scan the blob to reconstruct them since this will just result in the whole file being paged into memory, defeating the purpose of using `mmap`!)
2. Implement our own binary search (fine, that isn't hard to do)

At this point I'm pretty sure we don't need to do the "double bisection" algorithm anymore and can just search over the individual alignments.

dhdaines:
On further reflection implementing our own binary search isn't necessary (though it might be desirable).  The entry indices mentioned above can be wrapped in a class (should be a `NamedTuple` probably) which implements  `__le__` such that it looks at the binary blob to do comparisons.

PR #400, opened 2024-09-13
Dev.ej/compact lexicon

joanise:
### PR Goal? 

A bit of a crazy idea to reduce the amount of memory used by the g2p library: the English lexicon takes a lot of space in memory because each word is stored in a string, and each string object in Python is quite large, well beyond the size of the actual character sequence.

So, I thought, let's see what happens if I compact them by joining them by blocks. And a) it save 15MB in RAM from loading langs, b) with no measurable speed cost. The real cost is the increased complexity of the `find_alignment()` function in `g2p.mappings.utils`, but it goes from six lines to 15 lines, so, really, not that bad.

I've tested carefully to make sure the results are identical, and speed is not adversely affected.

The `langs.json.gz` file is a tad larger, but the memory footprint from loading it is significantly smaller.

### Fixes? 

Alternative solution to the problem raised in #395

### Feedback sought? 

Careful review.

### Priority? 

low

### Tests added? 

I better do that... coverage was not as good as I expected, and I will fix that to make sure I've tested all the corner cases.

### How to test? 

 - `g2p convert "some test" eng eng-ipa` still works as expected
 - unit tests & CI pass

### Confidence? 

high

### Version change? 

nope


dhdaines:
I think this actually *is* the memory mapping solution we're looking for, because if we put all the data together in a single (or multiple, whatever) block, then... we can easily memory map that block, without needing any external dependencies.

But also, if we support some kind of compact trainable G2P rules, we don't need to store the lexicon entries that are predictable by the rules, so that's something to think about (I could resume my debogosifying efforts on Phonetisaurus for instance)

joanise:
Before we use this as memory mapping, I'd want to convert the solution from splitting the compacted blocks into memory into reading the entries in place. But yeah, if we add the ability to access individual entries via string slices, we could turn the whole list into a single string, i.e., just one compact block.
For bisect to work on this, I guess we say `new_pos = (right + left)//2`, then scan back until we find `\x01`, from there scan forward to `\x01`, and that's the entry to examine.
I don't know if bisect supports anything but integral indexing -- probably not -- but it's a pretty darn simple algorithm to write ourselves -- that's one I've done a number of times -- so might be the solution indeed.

Question @dhdaines do we merge this anyway, take the 15mb saving, and plan a future update with a single block as described above, or do we hold off on this PR?
I vote for merging this PR in, because
 - it would already let us turn the ReadAlong-Studio Heroku server back down to a 1x from the current 2x.
 - the one-block solution will take a bit longer to write and I don't have time right now for it.

dhdaines:
> Question @dhdaines do we merge this anyway, take the 15mb saving, and plan a future update with a single block as described above, or do we hold off on this PR?

Yes, merge it now, because the internal representation of our rules/lexicons is not a public API/ABI so it really doesn't matter if we change it.

joanise:
> Yes, merge it now, because the internal representation of our rules/lexicons is not a public API/ABI so it really doesn't matter if we change it.

Sounds good.

😁  can you approve it then?

g2p/mappings/utils.py
```diff
@@ -495,18 +506,36 @@ def get_alignment_sequence(alignment: str, delimiter="") -> List[Tuple[int, str]
...
     """Given a sorted list of (word, alignment), find word and return its parsed alignment."""
     i = bisect_left(alignments, word)
-    if i != len(alignments):
-        k, v = alignments[i].split(_JOINER, maxsplit=1)
-        if k == word:
-            return get_alignment_sequence(v)
+    alignment_entry = _JOINER
+    if i != len(alignments) and alignments[i].startswith(word + _JOINER):
+        alignment_entry, _, _ = alignments[i].partition(_BLOCK_JOINER)
+    elif i > 0:
```
dhdaines, line 520: My first thought was ... but what if `i == 0 and not alignments[i].startswith(word + _JOINER)`?  But then I realized that this actually can't happen, since if `word > alignments[0]` then `bisect_left` will return 1.  It might be good to put an assertion in here with a comment to help future readers understand this.

g2p/mappings/utils.py
```diff
@@ -495,18 +506,36 @@ def get_alignment_sequence(alignment: str, delimiter="") -> List[Tuple[int, str]
...
+        alignment_block = alignments[i - 1].split(_BLOCK_JOINER)
+        j = bisect_left(alignment_block, word)
+        if j != len(alignment_block):
```
dhdaines, line 523: Likewise here ... I know it isn't possible that `j == 0` and the word *isn't* in the block, but it isn't obvious at first glance.

dhdaines (approved):
Approved, though you could add the assertions / comments mentioned if you want to make sure future generations understand what's going on :)

joanise:
Yes, adding that additional documentation (and possibly assertions) would be helpful, this is definitely not obvious code. Which is part of why I was as extensive as I was for unit testing, making unit testing confirm to me I didn't miss any corner cases.

joanise:
Thanks for the careful review and these comments.

joanise:
19be3945c0 documents the algorithm much more clearly now.

PR #405, opened 2024-10-31
Dev.ej/lexicon tokenizer

### LINKED ISSUE #401: Possible issue with tokenization
Here is the code I'm using in python:
```python
from g2p import make_g2p
TRANSDUCER = make_g2p('eng', 'eng-ipa')
test = TRANSDUCER("We'll let go, she'll stop.")
test.pretty_edges()
```
and the output:
```
[('W', 'w'),
 ('e', 'i'),
 ("'", "'"),
 (' ', ' '),
 ('l', 'l'),
 ('e', 'ɛ'),
 ('t', 't'),
 (' ', ' '),
 ('g', 'ɡ'),
 ('o', 'o'),
 ('o', 'ʊ'),
 (',', ','),
 (' ', ' '),
 ('s', 'ʃ'),
 ('h', 'ʃ'),
 ('e', 'i'),
 ("'", "'"),
 (' ', ' '),
 ('s', 's'),
 ('t', 't'),
 ('o', 'ɑ'),
 ('p', 'p'),
 ('.', '.')]
```

`We'll` and `she'll` might be tokenized as `we'` and `she'`. I'm not sure how or why this is happening or if this is understood behavior.  I'm not sure how to debug it.

joanise:
### PR Goal? 

Introduce the lexicon tokenizer

### Fixes? 

Fixes #401

### Feedback sought? 

This is a draft PR, seeking early feedback on the algorithm I implemented.

### Priority? 

medium-high

### Tests added? 

yes

### How to test? 

Run `g2p convert "We'll let go, she'll stop." eng eng-ipa` and see the output `wil lɛt ɡoʊ, ʃil stɑp.` instead of `wi' lɛt ɡoʊ, ʃi' stɑp.`

### Confidence? 

moderate only.

 - unit testing tells me it's fine
 - but I have not yet done any benchmarking to see if we need to optimize

### Version change? 

Probably yes, at least a patch bump, since this is a bug fix, maybe even a minor bump?

### DIFF
--- a/docs/package.md
+++ b/docs/package.md
@@ -32,10 +32,10 @@ Basic usage for the language-aware tokenizer:
 from g2p import make_tokenizer
 tokenizer = make_tokenizer("dan")
 for token in tokenizer.tokenize_text("Åh, hvordan har du det, Åbenrå?"):
-    if token["is_word"]:
-        word = token["text"]
+    if token.is_word
+        word = token.text
     else:
-        interword_punctuation_and_spaces = token["text"]
+        interword_punctuation_and_spaces = token.text
 ```
 
 Note that selecting the tokenizer language is important to make sure punctuation-like letters are handled correctly. For example `:` and `'` are punctuation in English but they will be part of the word tokens in Kanien'kéha (moh):
--- a/g2p/__init__.py
+++ b/g2p/__init__.py
@@ -16,10 +16,10 @@ Basic Usage:
     from g2p import make_tokenizer
     tokenizer = make_tokenizer(lang)
     for token in tokenizer.tokenize_text(input_text):
-        if token["is_word"]:
-            word = token["text"]
+        if token.is_word:
+            word = token.text
         else:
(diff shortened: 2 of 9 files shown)


joanise:
Some benchmarking notes: I tested g2p'ing the UNDHR (https://www.un.org/en/about-us/universal-declaration-of-human-rights) text, and it's actually about 1% faster on this branch than on `origin/main`, so clearly this implementation is fine on "normal" text.

I created a degenerate case with a word surrounded by dozens of punctuation marks on each side, and in that case I can see on slow-down on this branch, which makes sense because I'm doing O(n^2) lookups, when n is the number of candidate tokens, and each lookup costs O(log k), where k is the lexicon size. I'm incline to say YAGNI on optimizing this: it's only a problem on artificial data. Alternatively, I could cap the number of non-alphabetic characters before or after alphabetic ones that I include in the lexicon lookups, as a defensive programming solution.

PR #406, opened 2024-11-07
Dev.ej/refactor tokens

joanise:
### PR Goal? 

Define a proper class for tokens produced by the g2p tokenizer. Instead of a dict `{"text": "token_text", "is_word": True}`, we now have a `Token` class: `token = Token("token_text", True)`.

Access to the values can be done in the new and old ways:
 - `token.text` and `token.is_word` (recommended)
 - `token["text"]` and `token["is_word"]` (deprecated, but supported so this PR is not a breaking change)

Additional improvements:
 - ~~don't try to install pre-commit 3.2.0 on Python 3.7, it's not available (just don't develop g2p with 3.7, really!)~~ (moved to #408)
 - ~~quiet some noisy test suites~~ (moved to #408)
 - ~~black some things that should have been blacked before~~ (also #408)
 - add more typing, mostly required by the Token refactoring, but also going a little beyond that for things you can import directly `from g2p`, i.e., for g2p's programmatic API.

### Fixes? 

My annoyance as constantly manipulating these little dicts.

### Feedback sought? 

Sanity checking

### Priority? 

medium

### Tests added? 

yes

### How to test? 

Just run CI and see nothing is broken.

### Confidence? 

high

### Version change? 

This deserves a minor bump, and so does #405: they can be released together as 2.2.0.

### DIFF
--- a/docs/package.md
+++ b/docs/package.md
@@ -32,10 +32,10 @@ Basic usage for the language-aware tokenizer:
 from g2p import make_tokenizer
 tokenizer = make_tokenizer("dan")
 for token in tokenizer.tokenize_text("Åh, hvordan har du det, Åbenrå?"):
-    if token["is_word"]:
-        word = token["text"]
+    if token.is_word
+        word = token.text
     else:
-        interword_punctuation_and_spaces = token["text"]
+        interword_punctuation_and_spaces = token.text
 ```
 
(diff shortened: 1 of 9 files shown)


roedoejet (approved):
looks like a straight-forward refactor. I agree this is better than just a dict.

PR #409, opened 2024-11-15
Update the pretend scm version and make sure we don't forget it next time

joanise:
### PR Goal? 

Update the pretend scm version and make sure we don't forget it next time

### Fixes? 

tests fail locally now because `.SETUPTOOLS_SCM_PRETEND_VERSION` was not updated to the current major.minor version.

### Feedback sought? 

rubber stamping

### Priority? 

medium

### Tests added? 

 - `.SETUPTOOLS_SCM_PRETEND_VERSION`: already exercised in locally run unit testing

 - CI: cannot: I tested manually on my fork, to make sure the syntax is correct and passes on a match, and fails on a mismatch

### How to test? 

Don't waste your time, it requires removing all the real contents from `pythonpublish.yml`, but you can look at https://github.com/joanise/g2p/actions if you want to see my test results having done so.

### Confidence? 

high

### Version change? 

no


roedoejet (approved):
SCM had me looking for a .scm Scheme file 😅 - apart from the ambiguity as a result of me having spent too much time with Scheme, this looks good. Thanks for adding the extra comment in the release notes.

PR #410, opened 2024-12-17
fix: name ikt "Inuktut, Western" so it sorts next to Inuktitut

joanise:
### PR Goal? 

Make `ikt` sort next to `iku` and `iku-sro` in the RAS Web interface.

### Priority? 

low

### Confidence? 

high

### Version change? 

no, will just get picked up next time we release


roedoejet (approved):
looks good

PR #413, opened 2025-01-09
feat: add Halkomelem APA/Hul’q’umi’num’ mapping

goodzack:
Hul’q‛umi’num’ is the Island dialect of Halkomelem. This mapping is between APA used to transcribe Salish languages and the Hul’q’umi’num’ practical orthography.

Pull request template for adding a new language
-----------------------------------------------

* **Please check if the PR fulfills these requirements**
- [x ] Mapping files are added in `g2p/mappings/langs`
- [x ] Mapping is either added to an existing folder or a new folder has been added
- [x ] Language folder and files use appropriate [ISO 639-3 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-3_codes)
- [x ] `config-g2p.yaml` file includes all author names, and settings necessary
- [x ] Please add some test data in `g2p/tests/public/data`. The added file should be a csv/tsv/psv file and each row should have the format `[input_mapping_code,output_mapping_code,input_string,output_string]`
- [x ] As the last step, G2P has been updated by running `g2p update` locally and committing the change
- [x ] You agree to license your contribution under the same license as this project (see [LICENSE](https://github.com/roedoejet/g2p/blob/main/LICENSE) file).

* **Other information**:
Cheers 😛

PR #414, opened 2025-01-17
chore: remove obsolete network.pkl file

joanise:
Remove obsolete `network.pkl` file, accidentally re-introduced by #413


roedoejet (approved):
whoops! nice catch, thanks @joanise !

PR #416, opened 2025-03-06
ci: fix test-studio to work with Python 3.8

joanise:
### PR Goal? 

the latest playwright breaks CI for Python 3.8, so hold it back before the version that breaks it.


roedoejet (approved):
nicely done in between talks :) lgtm

PR #421, opened 2025-03-27
Fix several repo maintenance issues

joanise:
- update the pre-commit config and make it more standard
 - fix licensecheck -- needs Python >= 3.9 and adjusted configuration -- this one blocks other PRs since it makes CI fail
 - fix a mypy-detected error in app.py
 - matrix test more versions

### DIFF
--- a/.github/workflows/matrix-tests.yml
+++ b/.github/workflows/matrix-tests.yml
@@ -11,8 +11,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, windows-latest, macos-latest]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        os: [ubuntu-22.04, windows-latest, macos-latest]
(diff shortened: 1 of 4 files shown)


g2p/app.py
```diff
@@ -285,6 +285,7 @@ async def change_table(sid, message) -> None:
         mappings: List[Mapping] = []
         for lang1, lang2 in zip(path[:-1], path[1:]):
             transducer = make_g2p(lang1, lang2, tokenize=False)
+            assert isinstance(transducer, Transducer)
```
roedoejet, line 288: Why this assertion? I like putting error messages with assertion statements so that they're self-documenting, and also informative if the AssertionError gets raised. i.e. `assert isinstance(transducer, Transducer), "AssertionError raised because____"`

roedoejet (approved):
I've left this as approved, but I'd like to see the assert statement either removed or explained please - thanks!

g2p/app.py
joanise: Agreed, this naked assert is not really better than the stack dump it would pre-empt.

PR #425, opened 2025-05-12
build(deps): lock click<8.2.0 since it has breaking changes

joanise:
quick fix: click 8.2.0 has breaking changes we're not compatible with, most likely due to incompatibility with other dependencies, likely typer.

Look at the recent Actions runs of Matrix tests, that's where you'll see this patch making a difference, i.e., for Python >=3.10, since click 8.2.0 has dropped <3.10.


roedoejet (approved):
lgtm

PR #432, opened 2025-09-04
Add ces ipa - update

kubicra:

- [x ] Mapping files are added in `g2p/mappings/langs`
- [x ] Mapping is either added to an existing folder or a new folder has been added
- [ x] Language folder and files use appropriate [ISO 639-3 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-3_codes)
- [ x] `config-g2p.yaml` file includes all author names, and settings necessary
- [ x] Please add some test data in `g2p/tests/public/data`. The added file should be a csv/tsv/psv file and each row should have the format `[input_mapping_code,output_mapping_code,input_string,output_string]`
- [ x] As the last step, G2P has been updated by running `g2p update` locally and committing the change
- [ x] You agree to license your contribution under the same license as this project (see [LICENSE](https://github.com/roedoejet/g2p/blob/main/LICENSE) file).

* **Other information**:


kubicra:
Hello Eric,

I accidentally deleted my old branch so I had to do new pull request here. I copied your messages and my response from previous chat.

> > Dear Radek,
> 
> Thank you for submitting a PR with Czech mappings for the g2p package. And thank you for including test data!
> 
> There is a bug in your mapping, uncovered by your test data: rád gets mapped to raːd in IPA in some situations when it occurs in the middle of a phrase or sentence: by default, text gets tokenized before being mapped, but sometimes it does not, and in that case your contextual regex $ does not match the end of each word in Já mám rád čaj, only the end of the text.
> 
> You may have found the $ solution in our moh mapping, which actually has the same bug -- I will try to fix that.
> 
> The solution is unfortunately not as simple as we'd like: using $|  matches the end of string and when a word is followed by a space, but it would still fail to match before punctuation. For French I used \b because all French diacritics are considered word characters so having a string followed by the regex engine's definition of a word boundary reliably captures the end of a word. But that solution did not work for Czech because some diacritics in your test data appears to have been incorrectly considered non-word characters by the \b regular expression. I will have to investigate whether we have the same problem with moh or not.
> 
> It might take me a little while to figure out a good solution for this problem. It's been bugging me for some time now, I would like to find a stable solution for it!
> 
> Additional minor issues:
> 
> Our .csv file format does not accept a header line -- right now each of your .csv file has a mapping from in to out in the mappings database. They're harmless, because you also have left_context_regex as the left context and similar as the right context, so these rules would never match anything, but they're still noise in the database.
> 
> Your ces_pre and ces_pre_clean mappings don't do anything because the g2p module always uses the shortest path in the graph to convert a string to a multi-hop path. So it would always be ces->ces-ipa0->ces-ipa, never ces->ces0->ces->ces-ipa0->ces-ipa. If you need a pre-normalization step, we would normally create a ces -> ces-equiv mapping, and then provide only ces-equiv -> ces-ipa and no direct ces -> ces-ipa. Then, we you ask for ces to ces-ipa, the shortest path resolver will go through ces-equiv.
> 
> Finally, I would advise against stripping punctuation as you do in ces_pre_clean. The default mode for g2p is to tokenize text into word tokens and non-word tokens, and map only the characters in the word tokens. Now, the system infers that a character is part of a word by looking at the mapping chain: if you were to chain ces -> ces-clean (using ces_pre_clean) -> ces-equiv (using ces_pre) -> ces_ipa, you would end up declaring all those punctuations characters as word characters and I'm pretty sure that's not your intention.
> 
> So my advice is to use the following graph:
> 
> ces -> ces-equiv via ces_pre
> ces-equiv -> ces-ipa0 via ces_to_ces_ipa
> ces-ipa0 -> ces-ipa via ces_ipa_context, with corrections for word-final handling once I figure it out.
> Phonological question: does the word-final devoicing travel back to a preceding s, z, ɦ or r̝? You have rules saying d$ goes to t, sd goes to zd, and zt goes to st. But if a word ended with sd, should the result be zt or st? Right now you get zt because sd -> zd is applied first, then d$ -> t is applied. If devoicing should move back from the end of the word to a preceding instance of one of those consonants, you should put all the word final devoicing rules first in the mapping, so they are applied first.

> Some updates:
> 
> Looking for Czech text online, I did not find any examples of "z" or "s" before a word-final voiced stop, so maybe my question about regressive devoicing from the end of the word is not relevant? Please let me know if such words even exist, and if so, how they should be handled.
> 
> As for creating word-final rules, I think the correct solution is for us to warn that converting without tokenizing will break word-final transformations. The normal use of g2p is to tokenize first, and in that case $ works correctly to create word-final rules, just like you did. As long as that's how you intend to use it as well, things will work as expected. (Our ReadAlong Studio tokenizes first, for example, and g2p convert does so by default too.)

Hello Eric,

are your tests that you run internal or are those the same tests that I have run? Because I tried to put 'rád' to different parts of the sentence and with: "python run_tests.py langs" I get 'ra:t' always.

Our .csv file format does not accept a header line -- right now each of your .csv file has a mapping from in to out in >the mappings database. They're harmless, because you also have left_context_regex as the left context and similar as the >right context, so these rules would never match anything, but they're still noise in the database.**
Okay, I'll fix that

Your ces_pre and ces_pre_clean mappings don't do anything because the g2p module always uses the shortest path in >the graph to convert a string to a multi-hop path. So it would always be ces->ces-ipa0->ces-ipa, never ces->ces0->ces->ces->ipa0->ces-ipa. If you need a pre-normalization step, we would normally create a ces -> ces-equiv mapping, and then provide >only ces-equiv -> ces-ipa and no direct ces -> ces-ipa. Then, we you ask for ces to ces-ipa, the shortest path resolver will go >through ces-equiv.
Yeah, I must have overlooked that I used two different ces "nodes".

Finally, I would advise against stripping punctuation as you do in ces_pre_clean. The default mode for g2p is to tokenize >text into word tokens and non-word tokens, and map only the characters in the word tokens. Now, the system infers that a >character is part of a word by looking at the mapping chain: if you were to chain ces -> ces-clean (using ces_pre_clean) -> >ces-equiv (using ces_pre) -> ces_ipa, you would end up declaring all those punctuations characters as word characters and >I'm pretty sure that's not your intention.
Sure in the beginning I thought that is what should be done but you're right, it's not even done in practice.

So my advice is to use the following graph:

ces -> ces-equiv via ces_pre
ces-equiv -> ces-ipa0 via ces_to_ces_ipa
ces-ipa0 -> ces-ipa via ces_ipa_context, with corrections for word-final handling once I figure it out.
Yeah sounds reasonable

Phonological question: does the word-final devoicing travel back to a preceding s, z, ɦ or r̝? You have rules saying d$ >goes to t, sd goes to zd, and zt goes to st. But if a word ended with sd, should the result be zt or st? Right now you >get zt because sd -> zd is applied first, then d$ -> t is applied. If devoicing should move back from the end of the word >to a preceding instance of one of those consonants, you should put all the word final devoicing rules first in the mapping, so >they are applied first.

To answer that, when we pronounce things 'd$' is always 't', 'zt' goes always to 'st'. I think there is no word (or its modified form) that ends with 'sd' so this is okay.

Looking for Czech text online, I did not find any examples of "z" or "s" before a word-final voiced stop, so maybe my question >about regressive devoicing from the end of the word is not relevant? Please let me know if such words even exist, and if so, >how they should be handled.

There are some: for examlple 'újezd' or word 'mzda' (which means salary) has form 'mezd' according to one of the CASES rule in plural. As it is in both examples, it goes from 'zd' to 'st' and I am pretty sure, that it applies to not only these examples but to all cases.
And as I said I think there is literally 0 words that end with 'sd'. They only start with it and now when I think about that it is also impossible to encounter czech word which has 'sd' in the middle of the word 'blablaSDblabla'. In the word 'sdružit' which is 'to unify' 'sd' really goes to 'zd'.

PR #436, opened 2025-09-05
Dev.ej/tokenize in studio

### LINKED ISSUE #364: g2p-studio can't handle more than one word in English
In the CLI, running g2p on multiple words in English requires tokenization, which is now enabled by default:
```
$ g2p convert "astonishing, my friend" eng eng-arpabet
AH S T AA N IH SH IH NG , M AY  F R EH N D
$ g2p convert --no-tok "astonishing, my friend" eng eng-arpabet

$ g2p convert --no-tok "astonishing" eng eng-arpabet
AH S T AA N IH SH IH NG
```

In g2p studio, it looks like we are not enabling tokenization:
image
but
image

To align with our default behaviour on the CLI, I believe it would be best if the g2p studio also tokenized by default.

And wow, AH S T AA N... sounds so B AA S T AH N... Who wrote that pronunciation dict anyway??? ;)

joanise:
### PR Goal? 

While `g2p convert` tokenizes by default, and the read along studio tokenizes before mapping, g2p-studio did not do tokenization, and as a result we could not process more than one English word, and other mapping differences due to lack of tokenization would happen.

This PR turns on tokenization for the studio whenever it is possible to do so, which is when we're using existing mappings, though not when we're using a custom mapping (for which we can't know yet what tokenizer should be used).

Another bug with g2p-studio's handling of lexicon-based mappings was that we did not pass the parameters back to the back-end when we applied them. In particular, we did not pass the `case_sensitive=True` flag we set on eng->eng-ipa, so that `well` got mapped but `WELL` would end up blank.
At the same time, I'm enabling other options that can be customized at run time: if the user click on preserve-case, case-sensitive, changes the output separator or the norm form, they can see what would happen with such settings -- there is no reason not to allow the user to play with these settings interactively, especially when all we need to do is pass these flags to the back-end for them to work.

### Fixes? 

Fixes #364

Fixes potential discrepancies between g2p-studio and readalong-studio when using g2p-studio to debug text to submit to RAS.

Fixes the fact that lexicon mappings were always run case-sensitive despite their declarations

### Feedback sought? 

Good code review, and opinion on whether this is the right thing to do.

### Priority? 

low

### Tests added? 

hey, guess what! this is a great question to have here, because it made me stop and add them. So yes!

### How to test? 

Run `./run_studio.py` locally, chose eng as in_lang, and play with having one or more english words in lower or mixed case. Play with the mappings settings and see them change the output (when settings are coherent anyway).

### Confidence? 

medium-high

### Version change? 

Nothing more than a patch, if any.


g2p/app.py
```diff
@@ -213,6 +214,14 @@ async def convert(sid, message):
         )
         return
     transducer = CompositeTransducer(transducers)
+    try:
+        in_lang = message["data"]["mappings"][0]["kwargs"]["in_lang"]
+        tokenizer = make_tokenizer(in_lang)
+        transducer = TokenizingTransducer(transducer, tokenizer)
+    except Exception:
```
dhdaines, line 221: Someday, a linter may complain about this.  I see that there are a couple of distinct things that could go wrong in the try block, but as you say, tokenization is optional - the problem would arise if it starts to fail unexpectedly and we won't know exactly why.

dhdaines (approved):
Looks good (enough) to me!

PR #444, opened 2025-09-16
Refresh our production environment for Heroku

joanise:
### PR Goal? 

Dependanbot was complaining about starlette, and our production env has been left along for a while, so I figured it was a good idea to refresh it now. While there, I updated the readme with clearer instructions.

### Fixes? 

one dependabot alert

### Feedback sought? 

sanity checking

### Priority? 

low

### Tests added? 

already well tested in CI

### How to test? 

Look at the Heroku workflow in CI, or manually install the prod env and run `test_studio` and the other tests if you want to do it manually.

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/readme-heroku.md
+++ b/readme-heroku.md
@@ -1,13 +1,30 @@
+### Production environment on Heroku
+
 Our production Heroku deployment is controlled by the following files:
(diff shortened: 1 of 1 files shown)


roedoejet (approved):
looks good - thanks @joanise

PR #481, opened 2026-03-30
build(deps): rebuild Heroku reqs fresh to bump many dependencies

joanise:
### PR Goal? 

Refresh build dependencies for our Heroku deployment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,9 +39,9 @@ dependencies = [
   "pydantic>=2.4",
   "pyyaml>=5.2",
   "regex",
-  "text_unidecode",
+  "text-unidecode",
   "tqdm",
-  "typing_extensions",
+  "typing-extensions",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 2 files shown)


dhdaines (approved):
Seems totally fine to me.  I heard a rumor somewhere that Heroku will cease to exist in the near future, do we have a plan for this? (surely self-hosting is possible)

joanise:
> Seems totally fine to me. I heard a rumor somewhere that Heroku will cease to exist in the near future, do we have a plan for this? (surely self-hosting is possible)

Urgh, that would be a nuisance. I'll look into it. No plans yet, but thanks for bringing this to my attention!

We have a number of other apps depending on Heroku, so it's going to be high priority for me now.

joanise:
OK, looks like it's a slow wind-down: it should alive as is for the foreseeable future with no new features and no new corp customers. I see this as something to think about but not urgent, my expectation from their blog posts is that we might still have some years ahead, but without active development.

I think we can put moving to a different platform on the long-term todo list.

PR #483, opened 2026-04-02
fix(tests): use the default chromium channel for easier install

joanise:
### PR Goal? 

Use the default chromium channel from playwright instead of specifying the chrome channel. The latter breaks on my Ubuntu machine, while the former seems to work everywhere.

In particular, `playwright install chromium` can be done as a regular user, but `playwright install chrome` requires root access on Ubuntu, so the latter is easier to accomplish.

### How to test? 

@marctessier I have tested this on Windows and Linux, can you test it on Linux?

```
./run_studio.py &
g2p/tests/test_studio.py
```

should take about 30s to successfully run this suite.

### Confidence? 

high for Windows and Linux, hopeful for Mac too.

### Version change? 

no

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -67,7 +67,10 @@ jobs:
           key: g2p-neural-${{ steps.weekno.outputs.WEEK_NUMBER }}
           path: ~/.cache/huggingface
       - run: uv pip install -e .[neural]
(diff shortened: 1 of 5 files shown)


marctessier:
Look Good. after installing g2p  with ` pip install -e .[dev] `

and also `playwright install` afterwards... <-- only needed once

PR #489, opened 2026-04-24
Devise and test a heuristic to infer the IPA language code for a given lang

joanise:
### PR Goal? 

This work is triggered by EveryVoice currently not supporting `sal-apa` as an input language, because with get an `InvalidLanguageCode` exception from `sal-apa-ipa`. The correct IPA code is actually `sal-ipa`.

We don't have a rule at the moment deterministically saying how to derive the IPA language code from an input language code, but our convention happens to be either `lang_id+"-ipa"` or, if not found, `lang_id[:3]+"-ipa"`. The test case added in this PR formalizes that henceforth by asserting it to be so in unit testing.

### Fixes? 

While not actually fixing https://github.com/EveryVoiceTTS/EveryVoice/issues/789, this PR makes sure the solution I'm going to propose for that bug will keep working in the future.

### Feedback sought? 

careful analysis of `get_ipa_lang_code` in `test_langs.py`: do you have a better solution?

I would have preferred to add a proper function to g2p, but I want my solution to work on past and future versions of g2p, so that's not really possible. Instead, I'm formalizing via unit tests what I'm going to assume is true of the g2p library in EveryVoice.

### Priority? 

normal

### Tests added? 

yup

### How to test? 

    pytest g2p/tests/test_langs.py  -v

### Confidence? 

medium

### Version change? 

no

### DIFF
--- a/g2p/tests/test_langs.py
+++ b/g2p/tests/test_langs.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 
 import sys
-from unittest import TestCase
 
 from pytest import main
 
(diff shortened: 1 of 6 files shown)


g2p/tests/test_langs.py
```diff
@@ -1,49 +1,91 @@
...
+        output_string = transducer(test[2]).output_string.strip()
+        if output_string != test[3].strip():
+            LOGGER.error(
+                f"{error_prefix} for {test[-1]}: {test[2]} from {test[0]} to {test[1]} should be {test[3]}, got {output_string}"
+            )
+            error_count += 1
+
+    assert (
+        error_count == 0
+    ), f'g2p mapping errors found, look for "{error_prefix}" above for detail.'
```
joanise, line 41: `test_io()` above is not changed except for outdenting it from a `TestSuite` class method to a `pytest` style test function.

joanise:
@roedoejet PR updated. Note that critical review of this PR is important, because if we don't like `get_ipa_code()` some day, we won't get to change our mind about it given the promises I'm making here.

roedoejet (approved):
Looks good - thanks @joanise

PR #490, opened 2026-05-07
docs: replace pre-commit by prek in the contrib instructions

joanise:
### PR Goal? 

Replace pre-commit by prek https://prek.j178.dev/
 - prek uses uv to install hooks much faster
 - built in Rust, with the explicit intent to be fast
 - it can run the hooks in parallel when multiple files are affected
 - it's a drop-in replacement requiring no other change than doing `prek install` instead of `pre-commit install`

### Fixes? 

my growing annoyance with pre-commit, and really, just cause all these new rust-based tools actually are nicer. :) 

I've actually been using prek for months now, but I didn't to impose it on anyone. When I mentioned it to Samuel, he chided me  for now sharing the joy, so this PR is me finally doing that sharing.

It's not a huge difference, but it's still enough difference I found it worth switching, especially given how trivial it is to switch.

### How to test? 

Install `prek`:
 - redo `pip install -e ".[dev]", or just do `pip install prek`, or even `uv tool install prek` for a global install

Remove the pre-commit hook if you installed it before
```
rm .git/hooks/pre-commit
```

(Re-)install the hook:
```
prek install
```

Now commit some changes. `prek` will have to install everything in `.pre-commit-config.yaml` fresh, so the first time will be slow, though not nearly as slow as when using `pre-commit`, and subsequent times will be nice and fast.

### Confidence? 

high

### Version change? 

no

### Related PRs? 

https://github.com/EveryVoiceTTS/EveryVoice/pull/798
https://github.com/ReadAlongs/Studio/pull/303

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -29,14 +29,14 @@ install the required packages with `pip`:
 
 ```sh
 pip install -e ".[dev]"
-pre-commit install
+prek install
 gitlint install-hook
 ```
 
 ## Pre-commit hooks
 
 The g2p team has agreed to systematically use a number of pre-commit hooks to
-normalize formatting of code. You can install and enable pre-commit to have these used
+normalize formatting of code. You can install and enable prek or pre-commit to have these used
 automatically when you do your own commits.
 
 Pre-commit hooks enabled:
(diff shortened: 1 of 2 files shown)


roedoejet (approved):
lgtm, thanks for adding!

PR #492, opened 2026-06-05
build: drop Python 3.8 support and testing

joanise:
### PR Goal? 

Python 3.8 is long EOL, and it prevents us from using `uv lock` on Windows. There is no value in keeping it anymore, so let's drop it.

### Feedback sought? 

confirmation we don't need Python 3.8 support anymore

### Priority? 

normal

### Tests added? 

all well covered

### Confidence? 

high

### Version change? 

not required, but I would like to publish 2.3.2 soon with the recent dependency changes, this drop of 3.8, and the requirements.txt refresh in #493

### Related PR

https://github.com/ReadAlongs/Studio/pull/307

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: "3.8"
+          python-version: "3.10"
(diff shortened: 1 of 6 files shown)


roedoejet (approved):
Yes, I think it's fine to drop 3.8 now. thanks @joanise

PR #494, opened 2026-06-08
Schedule our test workflow to run weekly

joanise:
### PR Goal? 

Schedule our test workflow to run weekly.
 - refresh the cache as soon as the cache key rolls over
 - catch the effect of newly released versions of our dependencies early

Also address deprecation warning from unit tests: starlette 1.2+ wants httpx2 instead of httpx...

### Feedback sought? 

sanity

### Priority? 

normal

### How to test? 

To test this, we have to merge it first, then check next Monday morning to see if it ran...

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,8 +1,13 @@
 name: Run all tests
 on:
-  - pull_request
-  - push
-  - workflow_call
+  pull_request:
+  push:
+  workflow_call:
+  schedule:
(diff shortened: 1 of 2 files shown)


roedoejet (changes requested):
running at 1am UTC would give me notifications on Sunday night, can you please change it to run at 12pm/noon UTC? That would be 5am for me and 8am for you.

PR #496, opened 2026-07-01
feat: Add support of Polish language, and mapping to IPA, along with tests

wojcho:
* **Please check if the PR fulfills these requirements**
- [x] Mapping files are added in `g2p/mappings/langs`
- [x] Mapping is either added to an existing folder or a new folder has been added
- [x] Language folder and files use appropriate [ISO 639-3 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-3_codes)
- [x] `config-g2p.yaml` file includes all author names, and settings necessary
- [x] Please add some test data in `g2p/tests/public/data`. The added file should be a csv/tsv/psv file and each row should have the format `[input_mapping_code,output_mapping_code,input_string,output_string]`
- [x] As the last step, G2P has been updated by running `g2p update` locally and committing the change
- [x] You agree to license your contribution under the same license as this project (see [LICENSE](https://github.com/NRC-ILT/g2p/blob/main/LICENSE) file).

* **Other information**:

In [this PR](https://github.com/NRC-ILT/g2p/pull/431), it was indicated that `$` should not work as word boundary in context, although seemingly when testing, it worked correctly also outside of string end.

These tests illustrate this behavior:
```psv
pol|pol-ipa|mów|muf
pol|pol-ipa|mów głośniej|muf ɡwɔɕɲʲɛj
```

With this rule:
```json
{ "in": "w", "out": "f", "context_after": "$", "comment": "Word-final devoicing of <w>" },
```

If this solution has a mistake, I can modiffy the mapping.

PR #497, opened 2026-07-02
ci: check json.gz diff with unzipped contents, not compressed

joanise:
### PR Goal? 

When checking `mappings/langs/*.json.gz` to see if they are up to date, bit-for-bit-stable gzip is no longer possible since zlib-ng replaced zlib in Python 3.14 on Windows and will arrive sooner or later on all platforms. So we have to diff the unzipped contents, we can no longer rely on just the compressed contents.

See https://www.man.com/technology/faster-python for the a rationale for this change.

### Priority? 

high - partially interferes with #496 and there will be more like it in the future as the faster zlib-ng takes hold

### Tests added? 

n/a

### How to test? 

 - push a commit where the json.gz files are unchanged but compressed differently -- see CI pass (e.g., https://github.com/NRC-ILT/g2p/actions/runs/28613397570)
 - push a commit where the json.gz files are actually changed -- see CI fail (e.g., https://github.com/NRC-ILT/g2p/actions/runs/28613284328)

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -93,7 +93,7 @@ jobs:
           coverage combine
           coverage report
           coverage xml
-          if git status | grep -E 'static.*json|mapping.*pkl|mapping.*.json.gz'; then echo 'g2p databases out of date, please run "g2p update" and commit the results.'; false; else echo OK; fi
+
       - name: Upload coverage information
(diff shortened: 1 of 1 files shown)


roedoejet (approved):
Nice job catching this @joanise !

PR #503, opened 2026-07-08
fix: g2p update should not rewrite gzip files that are up to date

joanise:
### PR Goal? 

Stabilize `g2p update` output by not rewriting gzipped files that are up-to-date.

Related to #497, going one step further, so that if you get someone else's DB generated from a Python giving you different compressed output, you won't overwrite it (and thus recompress it) unless you've actually caused the file contents to change.

### Fixes? 

The risk of flip-flopping *.json.gz files as we each run tests and code stuff on different versions of Python and OSes.

### Priority? 

normal

### Tests added? 

~Thinking about it...~ yes

### How to test? 

`g2p update` will tell you that the two .json.gz files are up to date

```
gunzip g2p/mappings/langs/langs.json.gz
gzip -1 g2p/mappings/langs/langs.json
g2p update
```

will still tell you that the two .json.gz files are up to date, even though `git status` will say langs.json.gz has changed.

!!! Don't commit that change -- do `git restore g2p/mappings/langs/langs.json.gz` after to undo it.

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/g2p/mappings/langs/utils.py
+++ b/g2p/mappings/langs/utils.py
@@ -177,15 +177,31 @@ def cache_langs(
 
 
 def write_json_gz(path: str, data: Any):
+    """Write data into a .json.gz compressed file.
+
+    If the file exists but the contents have not changed, do not rewrite it: gzip output
+    is no longer stable with Python 3.14 starting to replace zlib by the faster zlib-ng,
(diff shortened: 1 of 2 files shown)


roedoejet (approved):
Tested - works well - thanks @joanise !

PR #504, opened 2026-07-23
simplify run_test and de-emphasize it

joanise:
### PR Goal? 

`run_tests.py` is mostly useless now that `pytest` is the standard way to run tests.
 - it gets stale whenever we add new test files
 - just running `pytest` is not harder! it's simpler

So:
 - adjust the documentation to tell people to use pytest, except for the one case where it kinda still has value: `./run_tests.py` skips the slow tests by default
 - strip all the pointless suites in run_tests
 - redefine `dev` subtractively so it auto-maintains itself: `dev = all - slow` where `slow` is explicitly defined instead
 - keep `run_tests.py langs` because the 7-part blog post mentions it, but still change `Contributing.md` to say `pytest g2p/tests/test_langs.py` instead, which is the recommended way now.
 - adjust the hatch configuration so you get a `dev` env by default: we'd never want to run `hatch shell` and not be able to run the tests.

### Fixes? 

removes stales stuff

### Feedback sought? 

nothing specific

### Priority? 

low

### How to test? 

- CI passes
- `pytest` runs all the tests
- `./run_tests.py --describe dev` confirms that `dev` runs all but the slow tests

### Confidence? 

high

### Version change? 

no

### DIFF
--- a/.github/workflows/matrix-tests.yml
+++ b/.github/workflows/matrix-tests.yml
@@ -29,7 +29,7 @@ jobs:
           python -m pip install --upgrade pip
           SETUPTOOLS_SCM_PRETEND_VERSION=$(cat .SETUPTOOLS_SCM_PRETEND_VERSION) pip install -e .[test]
       - name: Run tests
-        run: python run_tests.py dev
+        run: pytest
 
       # Neural tests, with huggingface models cached
       - name: Get current week number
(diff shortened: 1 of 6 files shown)


dhdaines:
Great!

roedoejet (approved):
lgtm - thanks @joanise-on-vacation!

PR #505, opened 2026-07-24
fix(kwk): use U+0315 instead of U+02BC in some places

roedoejet:
per Daisy Rosenblum: U+0315 is the Combining Comma Right, which is used for the glottal sounds U’mista represents as digraphs:

t̕s, t̕ł, k̕w, ḵ̕w


dhdaines (approved):
Ah, good catch. People don't necessarily use this consistently but it is the official usage, and what the keyboards will produce as well.

